Issue #7, opened 2018-03-18
Add tests


loic-sharma:
Today, the project is as simple as possible to enable fast iteration. Tests should be added once things stabilize.

WernerMairl:
Hi
I like the idea of this project!
I have started to play with this Issue in my own fork....
here first step:

https://github.com/WernerMairl/BaGet/commit/e5ed8cab908f981c9a837eb2d5e268934534aad9

please advice
- if PR's are welcome
- Rules/Guidelines to follow (my commit comment contains a few comments about my rule/guideline selection

regards
Werner

loic-sharma:
Replacing this with https://github.com/loic-sharma/BaGet/issues/480

PR #46, opened 2018-07-05
Add XUnit Infrastructure

WernerMairl:
# Xunit Infrastructure and Skeleton
First Steps done!
- 2 test assemblies created for 
    - 2 different implementation assemblies on 
        - 2 different layers (Controller and Core)

## BaGet.Core.Tests
First sample that allows testing Services

## BaGet.Tests
First sample that allows testing asp.net Controllers using a TestServer from Microsoft.AspNetCore.TestHost
Microsoft.Extensions.Logging redirected to Xunit.ITestoutputHelper

## Notes
- Naming guidelines like suggested in #7 

## Questions 
(to discuss - probably on another place)
### Test Infrastructure Code
now we have code like TestServerBuilder or Xunitlogger.
if this code base is growing we should create a separate assembly that contains all that code an can be references by multiple test-assemblies

### Test Data
we need some testdata - specially for unit tests.
if the test data can't be generated via code then a file based solution (resource files compiled into a assembly) are the most common/easiest solution
Resource Assemblies can be bigger => needs more time during compiling.
We should create a separate assembly for that. The content normally is not changed so incremental build (msbuild) can do his work.

### DIFF
--- a/src/BaGet.Services.Mirror/Services/FakeMirrorService.cs
+++ b/src/BaGet.Services.Mirror/Services/FakeMirrorService.cs
@@ -1,10 +1,20 @@
-﻿using System.Threading.Tasks;
+﻿using Microsoft.Extensions.Logging;
 using NuGet.Versioning;
+using System;
+using System.Threading.Tasks;
 
 namespace BaGet.Services.Mirror
 {
     public class FakeMirrorService : IMirrorService
     {
+
+        private readonly ILogger<MirrorService> _logger;
+
+        public FakeMirrorService(
+            ILogger<MirrorService> logger)
+        {
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
         public Task MirrorAsync(string id, NuGetVersion version) => Task.CompletedTask;
     }
 }
--- a/src/BaGet.Services.Mirror/Services/FakeMirrorService.cs
+++ b/src/BaGet.Services.Mirror/Services/FakeMirrorService.cs
@@ -1,20 +1,10 @@
-﻿using Microsoft.Extensions.Logging;
-using NuGet.Versioning;
-using System;
+﻿using NuGet.Versioning;
 using System.Threading.Tasks;
 
 namespace BaGet.Services.Mirror
 {
     public class FakeMirrorService : IMirrorService
     {
-
-        private readonly ILogger<MirrorService> _logger;
-
-        public FakeMirrorService(
-            ILogger<MirrorService> logger)
-        {
-            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
-        }
         public Task MirrorAsync(string id, NuGetVersion version) => Task.CompletedTask;
     }
 }
--- a/src/BaGet.Services.Mirror/Services/FakeMirrorService.cs
+++ b/src/BaGet.Services.Mirror/Services/FakeMirrorService.cs
@@ -1,5 +1,5 @@
-﻿using NuGet.Versioning;
-using System.Threading.Tasks;
+﻿using System.Threading.Tasks;
+using NuGet.Versioning;
 
 namespace BaGet.Services.Mirror
 {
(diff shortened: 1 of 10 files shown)


src/BaGet.Services.Mirror/Services/FakeMirrorService.cs
```diff
@@ -1,10 +1,20 @@
...
 {
     public class FakeMirrorService : IMirrorService
     {
+
+        private readonly ILogger<MirrorService> _logger;
+
+        public FakeMirrorService(
+            ILogger<MirrorService> logger)
+        {
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
```
loic-sharma: What's this logger for?

src/tests/BaGet.Core.Tests/FilePackageStorageServiceTests.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    public class FilePackageStorageServiceTests
+    {
+        [Fact]
+        public void CreatesStoreDirectoryIfItDoesNotExist()
+        {
+            string validButNotExistingPath = Path.Combine(Path.GetTempPath(), System.Guid.NewGuid().ToString("N"));
+            Assert.False(Directory.Exists(validButNotExistingPath));
+            IPackageStorageService service = new FilePackageStorageService(validButNotExistingPath);
+        }
```
loic-sharma: Should this have an `Assert.False(Directory.Exists(validButNotExistingPath));` after instantiation?

src/BaGet.Services.Mirror/Services/FakeMirrorService.cs
WernerMairl: good question:
not commited the one dummy test message to the logger, that i have used to verify that the output goes to ITestoutputhelper

src/tests/BaGet.Core.Tests/FilePackageStorageServiceTests.cs
WernerMairl: currently this assertion throws a exception, because the constructor does NOT create the directoy (created later on demand)

should we create/ensure the directory during FilePackageStorageService construction ?

if we follow „fail fast and early“ probably yes....

if we do things like that inside the constructor:

-creating the directory verifies write permission.... nice
-should we do another check of write permission in case of directoryexists ?

src/BaGet.Services.Mirror/Services/FakeMirrorService.cs
WernerMairl: should i
a) remove the logger argument
b) add a logger output to the Fakeserver

it is not a big difference ...if i hear nothing from your side i do 
b) 
as next fix for this PR

src/BaGet.Services.Mirror/Services/FakeMirrorService.cs
```diff
@@ -1,5 +1,5 @@
-﻿using System.Threading.Tasks;
-using NuGet.Versioning;
+﻿using NuGet.Versioning;
+using System.Threading.Tasks;
```
loic-sharma: Please keep `System` namespaces at the top (it's the .NET team's style)

src/tests/BaGet.Tests/PackageControllerTest.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            Helper = helper ?? throw new ArgumentNullException(nameof(helper));
+        }
+
+        [Theory]
+        [InlineData("id01")]
+        [InlineData("id02")]
+        public async Task AskEmptyServerForNotExistingPackageID(string packageID)
+        {
+            using (TestServer server = TestServerBuilder.Empty().TraceToTestOutputHelper(this.Helper,LogLevel.Error).Build())
```
loic-sharma: No need for `this.` here

loic-sharma:
Okay, finally reviewed it. Looks great! I left some minor nitpicks, I'll go ahead and merge it once it's fixed. Thank you! 😄

src/BaGet.Services.Mirror/Services/FakeMirrorService.cs
WernerMairl: done!

src/tests/BaGet.Tests/PackageControllerTest.cs
WernerMairl: done!

PR #59, opened 2018-07-12
Added dockerfile

### LINKED ISSUE #37: Create a docker image
/cc @emgarten

Baklap4:
### What does this PR do?
Adds two dockerfiles. One is for deploy the other is for development.

### Closes Issue(s)

Closes #37 

### Motivation
Dockerizing an app is fun to do and detaches all host dependencies making it easier to upgrade the application.

### DIFF
--- a/src/Dockerfile.dev
+++ b/src/Dockerfile.dev
@@ -1,0 +1,15 @@
+# Build environment
+FROM microsoft/dotnet:2.1.302-sdk AS build-env
+WORKDIR /src
+COPY * ./
+RUN dotnet restore && dotnet publish -c Debug -o out
+
+# Build runtime image
+FROM microsoft/dotnet:2.1.1-aspnetcore-runtime-alpine
+ENV ASPNETCORE_ENVIRONMENT Development
+ENV ASPNETCORE_URLS http://0.0.0.0:5000
+EXPOSE 5000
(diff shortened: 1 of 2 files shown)


Baklap4:
@loic-sharma if i'm missing some build step please say i'll add them to it

src/Dockerfile.dev
```diff
@@ -0,0 +1,16 @@
+FROM microsoft/dotnet:2.1.302-sdk
```
loic-sharma: Is it common to use docker during development? I personally don't, it seems like an extra hoop. If so, let's stick to one Dockerfile.

WernerMairl: @Baklap4 multi stage build (one stage on a dev image and the final stage on a runtime image ?

Baklap4: @loic-sharma It's very common to use docker during development, as this will make sure if it works on your computer it works on everyone else's

@WernerMairl thanks for the suggestion will look into multistage, this would deffo reduce the build of the `Dockerfile.dev` and shrink it to 60mb (due to the underlying base image dotnet-runtime-xxx

loic-sharma:
Hey, thanks for your contribution! A dockerfile was added by this change: https://github.com/loic-sharma/BaGet/pull/61

PR #64, opened 2018-07-31
write file to disk if it doesn't exist

smazurov:
### What does this PR do?

If a file is in the db, but not on disk, currently registration fails with 409, without updating file on disk. This PR re-saves file to disk (if for instance, it was deleted or not persisted)

### Motivation

As I was setting it up to host our internal nuget packages, I kept running conflict issues when my files weren't even there. My motivation was to harden the implementation a little bit, without changing any of the client expectations (it will still return a 409, for instance)

### DIFF
--- a/src/BaGet.Core/Services/IndexingService.cs
+++ b/src/BaGet.Core/Services/IndexingService.cs
@@ -7,10 +7,12 @@ using BaGet.Core.Entities;
 using BaGet.Core.Extensions;
 using Microsoft.Extensions.Logging;
 using NuGet.Packaging;
+using NuGet.Packaging.Core;
+using PackageDependency = BaGet.Core.Entities.PackageDependency;
 
 namespace BaGet.Core.Services
 {
-    using BaGetPackageDependency = Entities.PackageDependency;
+    using BaGetPackageDependency = PackageDependency;
 
     public class IndexingService : IIndexingService
     {
@@ -44,7 +46,9 @@ namespace BaGet.Core.Services
                     var packageId = packageReader.NuspecReader.GetId();
                     var packageVersion = packageReader.NuspecReader.GetVersion();
 
-                    if (await _packages.ExistsAsync(packageId, packageVersion))
+                    var identity = new PackageIdentity(packageId, packageVersion);
+
+                    if (await _storage.ExistsAsync(identity))
                     {
                         return IndexingResult.PackageAlreadyExists;
                     }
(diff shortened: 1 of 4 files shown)


smazurov:
Yup, exactly. More specifically it was a persistent volume claim that was getting nuked. I realize that the chances of it happening are low but I wrote some code that fixes it so I figured it’s useful. I can see a case where you switch storage providers and don’t want to reset package stats but need the file to exist

src/BaGet.Core/Services/IndexingService.cs
```diff
@@ -44,7 +46,9 @@ public class IndexingService : IIndexingService
                     var packageId = packageReader.NuspecReader.GetId();
                     var packageVersion = packageReader.NuspecReader.GetVersion();
 
-                    if (await _packages.ExistsAsync(packageId, packageVersion))
+                    var identity = new PackageIdentity(packageId, packageVersion);
+
+                    if (await _storage.ExistsAsync(identity))
```
loic-sharma: We should not fail uploads if the package already exists in storage. Consider the following scenario:

1. A customer starts a package upload
2. The package is uploaded to storage
3. BaGet throws an exception before the package's metadata is persisted to the database

If we fail uploads when the package already exists in storage, we are now in an unrecoverable state. An admin needs to go manually delete the package from storage. The solution to these problems is to make the database the single source of truth.

loic-sharma:
I don't think this feature is a good idea due to the concerns I raised in my previous comment. I think a better alternative would be be to provide tooling to clean up weird state.

smazurov:
I typically would agree, but in this case you can't tool around missing files besides blowing things away. I slightly changed the implementation that addresses your concern.

loic-sharma:
I'm still not sold on this feature. I see the value in being able to fix weird state, however, I think that getting this "fix up" behavior right is difficult. For example, what if the package changed when it was reuploaded? Ideally, the database's metadata and the search service's index should both be updated.

> I can see a case where you switch storage providers and don’t want to reset package stats but need the file to exist

I agree that a way to switch storage providers would be useful. However, switching storage providers by running `nuget push` on each stored package would be a poor experience. I think the right solution would be a dedicated tool to "mirror" packages from one storage provider to another.

Really, customers shouldn't need to do storage "fix ups" unless they fiddled around with their state by hand. I think we'd be better off providing tools that do whatever manual actions customers want to do. What do you think?

smazurov:
Its your call, the way I see it is:

1) its useful to have `ExistsAsync` task in storage interface and implementations
2) it doesn't hurt anything, it simply makes the implementation that exists already a bit more robust.
3) changes you propose are great, but obviously outside the scope of this. 

I won't be offended if you just close it and not accept it.

loic-sharma:
Fair enough. I'll close this for now, but we can revisit this feature if it becomes a common ask. Thank you for your contributions!

PR #65, opened 2018-07-31
add allowed forwarded headers so it can work behind proxy

smazurov:
### What does this PR do?

Allows BaGet to work behind an https proxy

### Closes Issue(s)
none

### Motivation

We have ssl termination at our proxy, with BaGet running behind it over http. Before the urls were all wrong (with `http` protocol), after this, they're all correct

### Additional Notes

I decided to let aspnet trust all proxies, but in the future it should probably be configurable.

### DIFF
--- a/src/BaGet/Startup.cs
+++ b/src/BaGet/Startup.cs
@@ -10,6 +10,7 @@ using BaGet.Extensions;
 using BaGet.Web.Extensions;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.HttpOverrides;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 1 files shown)


src/BaGet/Startup.cs
```diff
@@ -45,6 +46,14 @@ public void ConfigureServices(IServiceCollection services)
                         .AllowAnyMethod()
                         .AllowAnyHeader());
             });
+            
+            services.Configure<ForwardedHeadersOptions>(options =>
+            {
+                options.ForwardedHeaders = ForwardedHeaders.XForwardedFor | ForwardedHeaders.XForwardedProto;
+                // do not restrict to local network/proxy
```
loic-sharma: Super nitpick: please capitalize your comment 😄

loic-sharma:
Looks great! I left a minor comment on style, I'll merge this in once that's fixed!

smazurov:
done, also fixed my gpg, oops.

loic-sharma:
Sweet, thank you! 🥂

Issue #73, opened 2018-08-04
Split up the BaGet.Web project


loic-sharma:
Right now, `BaGet.Web` is a netcoreapp2.1 project as it references `Microsoft.AspNetCore.App`. It should be split into two pieces:

1. Anything related to the serialization of NuGet's v3 protocol should be moved to a `BaGet.Protocol` project. This project should be netstandard2.0
2. The rest should be moved to the `BaGet` project.

WernerMairl:
some additional information about ".AddApplicationPart", reason and history...
(we reference this issue in a comment inside a cs file)

https://github.com/aspnet/Mvc/issues/5992
https://stackoverflow.com/questions/43669633/why-is-testserver-not-able-to-find-controllers-when-controller-is-in-separate-as?noredirect=1#comment74386164_43669633

loic-sharma:
This work is done.

PR #81, opened 2018-08-07
Fix Test Infrastructure

WernerMairl:
### What does this PR do?
Fixes problems with false positive tests

Changes: 
- call "AddApplicationPart()" for MVC => it helps MVC to find additional assemblies!
- a first test for Searchcontroller
- add some missing/broken config option!

### DIFF
--- a/src/BaGet/Extensions/ServiceCollectionExtensions.cs
+++ b/src/BaGet/Extensions/ServiceCollectionExtensions.cs
@@ -99,7 +99,7 @@ namespace BaGet.Extensions
 
         public static IServiceCollection ConfigureHttpServices(this IServiceCollection services)
         {
-            services.AddMvc();
+            services.AddMvc().AddApplicationPart(typeof(BaGet.Web.Routes).Assembly);
             services.AddCors();
             services.AddSingleton<IConfigureOptions<CorsOptions>, ConfigureCorsOptions>();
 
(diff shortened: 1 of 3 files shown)


src/BaGet/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -99,7 +99,7 @@ public static IServiceCollection AddBaGetContext(this IServiceCollection service
 
         public static IServiceCollection ConfigureHttpServices(this IServiceCollection services)
         {
-            services.AddMvc();
+            services.AddMvc().AddApplicationPart(typeof(BaGet.Web.Routes).Assembly);
```
loic-sharma: I haven't had any issues loading controllers from `BaGet.Web`. How did you run into these issues?

Also, I'm considering moving the controllers back to the `BaGet` assembly and renaming `BaGet.Web` to `BaGet.Protocol`. What do you think? The discussion for that is here: https://github.com/loic-sharma/BaGet/issues/73

loic-sharma: Could you add a comment like:

```
// TODO: Delete this
// See: https://github.com/loic-sharma/BaGet/issues/73
```

WernerMairl: comment added and added a few informations (links) to 73

tomzo: I just ran into this. Definitely needed.
@loic-sharma please merge this. It just adds a test and fixes assembly loading so no point in waiting.

tomzo:
@WernerMairl maybe you could rebase on master.

loic-sharma:
The controllers were brought back into the BaGet assembly, so this change is no longer needed. Thanks for opening this @WernerMairl!

PR #82, opened 2018-08-07
[WIP] Discussion "Options Validation"

WernerMairl:
first draft (one sample) - we can use GitHub Conversation feature for discussion.

Changes: 
- it avoids putting methods on the "Option" classes....
- it uses Extensions Methods BUT the this object is not the (may be nullable) Option class but the IServiceProvider (not nullable)

### DIFF
--- a/src/BaGet/Extensions/ServiceCollectionExtensions.cs
+++ b/src/BaGet/Extensions/ServiceCollectionExtensions.cs
@@ -22,6 +22,56 @@ using Microsoft.Extensions.Options;
 
 namespace BaGet.Extensions
 {
+
+    public static class ServiceProviderExtensions
+    {
+
+        public static DatabaseOptions GetDatabaseOptions(this IServiceProvider serviceProvider)
+        {
+            if (serviceProvider == null) throw new ArgumentNullException(nameof(serviceProvider));
+
+            var databaseOptions = serviceProvider.GetRequiredService<IOptions<BaGetOptions>>()
+                                    .Value
+                                    .Database ?? throw new ArgumentNullException(nameof(serviceProvider));
+
+            if (databaseOptions == null) BaGet.Core.Extensions.OptionsExtensions.ThrowMissingConfiguration(nameof(BaGetOptions.Database));
+
+            if (string.IsNullOrEmpty(databaseOptions.ConnectionString))
+            {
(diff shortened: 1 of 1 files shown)


src/BaGet/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -22,6 +22,56 @@
 
 namespace BaGet.Extensions
 {
+
+    public static class ServiceProviderExtensions
+    {
+
+        public static DatabaseOptions GetDatabaseOptions(this IServiceProvider serviceProvider)
```
loic-sharma: Can we get rid of the empty line on line 28?

src/BaGet/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -22,6 +22,56 @@
...
+    public static class ServiceProviderExtensions
+    {
+
+        public static DatabaseOptions GetDatabaseOptions(this IServiceProvider serviceProvider)
+        {
+            if (serviceProvider == null) throw new ArgumentNullException(nameof(serviceProvider));
+
+            var databaseOptions = serviceProvider.GetRequiredService<IOptions<BaGetOptions>>()
+                                    .Value
+                                    .Database ?? throw new ArgumentNullException(nameof(serviceProvider));
```
loic-sharma: ` .Database ?? throw new ArgumentNullException(nameof(serviceProvider));` should be ` .Database;` as you check that `databaseOptions` isn't null on line 37.

src/BaGet/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -22,6 +22,56 @@
...
+                    nameof(BaGetOptions.Database),
+                    nameof(DatabaseOptions.ConnectionString));
+            }
+
+            return databaseOptions;
+        }
+
+
+
+        public static IContext GetDatabaseContext(this IServiceProvider serviceProvider)
```
loic-sharma: Could we compress all the empty lines on lines 48-50 into a single line?

loic-sharma: Also, please remove all the empty lines from lines 70-74.

loic-sharma:
Hey, thanks for starting this work @WernerMairl. This was finally addressed by https://github.com/loic-sharma/BaGet/pull/147

Issue #88, opened 2018-09-03
Request body too large


rong0312:
Greetings,

After integration and testing, I found that certain package pushes got response code of 500,
and this is the reason: 

![screen shot 2018-09-03 at 14 19 26](https://user-images.githubusercontent.com/33253167/44984140-62077580-af84-11e8-9edf-d4d9ca47b4bc.png)

The packages that didn't make the push (and the size):

![screen shot 2018-09-03 at 14 23 33](https://user-images.githubusercontent.com/33253167/44984323-fc67b900-af84-11e8-89b6-cfb6b78d4f1c.png)

Seems that BaGet wont push packages over a given size.
Is this issue known? 
Is there any solution / possibility to increase the allowed post request?
This is the only issue that prevents our deployment on production.  

Thanks a lot, Ron

rong0312:
UPDATE;

I solved this issue, I made a PR regards this solution.

loic-sharma:
Fixed by https://github.com/loic-sharma/BaGet/commit/f3adb30bb2028673bfc78d960ad234c61e2d48dd

PR #89, opened 2018-09-03
Update Program.cs & adding import script (fix "Request body too large" error).

rong0312:
### What does this PR do?

First issue is about  fixing the "body too large" error when posting to BaGet repo. 
(push that exceeds 27 MB)

More over, another addition is python script, "pushPackage.py", that is meant to allow automatic
import of existed NuGet repository into BaGet repository. 
The script has parameters who have to be altered ( "<XYZ> syntax"). 
"failedPushes" is variable that presents directory and will serve to handle failed (status code != 200).
All failed packages will be located there for the programer's QA and fixes. 

### Closes Issue(s)
Addresses https://github.com/loic-sharma/BaGet/issues/88

### Motivation

I had this issue.
I had to import 300 packages from existing repository, that motivated me to create the script.

### DIFF
--- a/pushPackage.py
+++ b/pushPackage.py
@@ -1,0 +1,29 @@
+import os
+import subprocess
+import shutil
+
+packagesPath = "/Users/ron/dev/python/content"
+failedPushes = "/Users/ron/dev/python/failed"
+uploadCount = 0
+failedCount = 0
+contentDir = os.listdir(packagesPath)
+
+for package in contentDir:
+    pushCommand = "dotnet nuget push %s -k ECB-NUGET-SERVER-KEY -s http://%s:80/v3/index.json" % (package, os.getenv('PRIMARY_IP_ADDRESS'))
+    result = subprocess.call(['bash','-c', "cd %s && %s" % (packagesPath, pushCommand)]) 
+    if result is not 0:
+        try:
+            shutil.move("%s/%s" % (packagesPath, package), "%s/%s" % (failedPushes, package))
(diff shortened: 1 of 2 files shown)


src/BaGet/Program.cs
```diff
@@ -45,7 +45,11 @@ await provider
 
         public static IWebHostBuilder CreateWebHostBuilder(string[] args) =>
             WebHost.CreateDefaultBuilder(args)
-                .UseStartup<Startup>();
+                .UseStartup<Startup>()
+                .UseKestrel(options =>
+                {
+                    options.Limits.MaxRequestBodySize = null;
```
loic-sharma: Let's set `MaxRequestBodySize` to `262_144_000` (250 MB) to match nuget.org's upload limit (see [this documentation](https://docs.microsoft.com/en-us/nuget/policies/nuget-faq#managing-packages-on-nugetorg)).

pushPackage.py
```diff
@@ -0,0 +1,29 @@
+import os
+import subprocess
+import shutil
+
+packagesPath = "/path/to/existed/package/directory"
```
loic-sharma: I'd rather make this script an "official" command on the BaGet command line. Something like `baget import /path/to/local/feed`.

Would you mind removing the script from this change so that we can fix the upload limit right away? 😄

loic-sharma:
Thanks for finding the fix. I left some comments, please address them and then I'll merge this in!

PR #99, opened 2018-09-14
Fixed docker build. Added option to configure build type

RKalkani:
### What does this PR do?

- PR added option to default run release build. The user can set the APP_BUILD argument during docker image build to have a Debug build.
- Renamed dockerfile to default standard name "Dockerfile".
- Fix docker build


loic-sharma:
Thank you!

Issue #93, opened 2018-09-06
Caching question


rong0312:
Question issue, regards caching;

I'v noticed this feature:

![isue](https://user-images.githubusercontent.com/33253167/45165233-098edd00-b1fd-11e8-84d8-2d4d65199323.png)

I have enabled it.

However, when i install package from nuget.org i cant see it in my BaGet volume. 
How can i tell this feature works?

thanks a lot!

loic-sharma:
> However, when i install package from nuget.org i cant see it in my BaGet volume.

This is odd - you should see the cached nupkg in your BaGet volume. Is the volume empty?

rong0312:
>This is odd - you should see the cached nupkg in your BaGet volume. Is the volume empty?

Unfortunately the package didn't make it, 
I am installing packages in VS from nuget.org

loic-sharma:
What NuGet sources are you using in Visual Studio? You should have just `http://localhost:5555/v3/index.jon` to force Visual Studio to use BaGet. If you have both nuget.org and BaGet, Visual Studio may use nuget.org to download the package.

Also, have you already installed this package? NuGet won't contact upstream feeds if it has already downloaded the package. Check your machine's offline mirrors (like `C:\Users\your-usernam\.nuget\packages`) to see if the package has already.

rong0312:
I removed the nuget.org source and tried to install package from there via the console:

![screen shot 2018-09-17 at 14 25 41](https://user-images.githubusercontent.com/33253167/45620642-41601500-ba86-11e8-9f8e-74df885744ff.png)

Did you come across something like this?

tomzo:
I took a look on what happens here.
I have started baget under debugger. Then ran
```
nuget install log4net -DisableParallelProcessing -NoCache -Source http://localhost:50561/v3/index.json
```
This will hit `RegistrationIndexController` which calls `IPackageService` and returns NotFound.
```
var packages = await _packages.FindAsync(id);
            var versions = packages.Select(p => p.Version).ToList();

            if (!packages.Any())
            {
                return NotFound();
            }
```
@loic-sharma how was this supposed to work? isn't the package service responsible only for indexing local packages?
Would you accept a PR in which cache is moved to separate endpoint? E.g. `v3/cache/index.json`?

loic-sharma:
Apologies, I haven't had a chance to look into this bug yet.

> isn't the package service responsible only for indexing local packages

[...]

PR #102, opened 2018-09-28
WIP fix/implement caching upstream #93

tomzo:
For read-through caching, queries to registration and package service
must be directed to upstream server. Otherwise client would get always
404 because there is nothing in the cache to begin with.
We can add caching queries too, but that requires integrating catalog
reader to refresh responses for already cached packages.

### What does this PR do?

Fixes read-through cache.

### Closes Issue(s)

#93

### Additional Notes

 * still need to add an option to parametrrize cache source. I don't think previous option was ever sufficient `"PackageSource": "https://api.nuget.org/v3-flatcontainer/"`. I think we need to specify full URL to the index of nuget server. Such as `"https://api.nuget.org/v3/index.json"`.
 * this is working now, but there are 3 endpoints which work by checking local packages first and then fallback to ask nuget.org. I think it would be more sensible if caching endpoint was separate from private index. Please tell me if I can move it.
 * I'll add unit tests before merge.

### DIFF
--- a/src/BaGet.Core/Mirror/FakeMirrorService.cs
+++ b/src/BaGet.Core/Mirror/FakeMirrorService.cs
@@ -1,5 +1,7 @@
-﻿using System.Threading;
+﻿using System.Collections.Generic;
+using System.Threading;
 using System.Threading.Tasks;
+using NuGet.Protocol.Core.Types;
 using NuGet.Versioning;
 
 namespace BaGet.Core.Mirror
@@ -9,6 +11,19 @@ namespace BaGet.Core.Mirror
     /// </summary>
     public class FakeMirrorService : IMirrorService
     {
+        Task<IReadOnlyList<string>> emptyVersions = Task.Factory.StartNew(() => new List<string>() as IReadOnlyList<string>);
+        Task<IEnumerable<IPackageSearchMetadata>> emptyMeta = Task.Factory.StartNew(() => new List<IPackageSearchMetadata>() as IEnumerable<IPackageSearchMetadata>);
+
+        public Task<IReadOnlyList<string>> FindUpstreamAsync(string id, CancellationToken ct)
+        {
+            return emptyVersions;
+        }
+
+        public Task<IEnumerable<IPackageSearchMetadata>> FindUpstreamMetadataAsync(string id, CancellationToken ct)
+        {
+            return emptyMeta;
+        }
+
         public Task MirrorAsync(
             string id,
             NuGetVersion version,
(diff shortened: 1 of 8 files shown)


src/BaGet.Core/Mirror/FakeMirrorService.cs
```diff
@@ -9,6 +11,19 @@ namespace BaGet.Core.Mirror
     /// </summary>
     public class FakeMirrorService : IMirrorService
     {
+        Task<IReadOnlyList<string>> emptyVersions = Task.Factory.StartNew(() => new List<string>() as IReadOnlyList<string>);
```
loic-sharma: Instead of `Task.Factory.StartNew`, how about `Task.FromResult`?

loic-sharma:
Thank you for tackling this! I think your current solution has a flaw. Say package `Test` has two versions: v1.0.0 and v2.0.0. This behavior seems confusing:

1. I setup a clean installation of BaGet
2. I browse `v3/registration/test/index.json` and see both versions as expected
3. I browse `v3/registration/test/2.0.0.json`, package 2.0.0 gets mirrored
4. I browse `v3/registration/test/index.json` and see just version 2.0.0. Version 1.0.0 is missing!

This is a tough problem to solve. What're your thoughts @tomzo?

src/BaGet.Web/Controllers/Registration/RegistrationIndexController.cs
```diff
@@ -141,6 +172,41 @@ public CatalogEntry(Package package, string catalogUri, string packageContent)
...
+
+                PackageId = package.Identity.Id;
+                Version = package.Identity.Version.ToFullString();
+                Authors = string.Join(", ", package.Authors);
+                Description = package.Description;
+                Downloads = package.DownloadCount.GetValueOrDefault(0);
+                HasReadme = false; // 
+                IconUrl = NullSafeToString(package.IconUrl);
+                Language = null; //
+                LicenseUrl = NullSafeToString(package.LicenseUrl);
```
loic-sharma: Instead of a helper, I would do just `package.LicenseUrl?.ToString()`

src/BaGet.Core/Mirror/FakeMirrorService.cs
tomzo: Will do.

tomzo:
> This is a tough problem to solve. What're your thoughts @tomzo?

Second endpoint would solve that too.
Because currently this happened:
 * in `1.` query to db returned none, so we fallback to mirror.
 * in `4.` you got response based on non-empty reply from query to db, which now contains package which is treated same as private one.
If we don't split private/public cached packages, that each reply will have to also check with upstream to see what other packages are available.

tomzo:
I have implemented caching with new endpoint in [my fork](https://github.com/ai-traders/baget), which solves above problem. Cached packages are not indexed in database at all, we just query nuget.org.
I can submit a PR once #108 one is merged.

PR #104, opened 2018-10-01
Fix index versions and urls; add tests

tomzo:
`/Versioned` looks like a leftover and this is definitely not the right version in any of the endpoints.
These changes were enough to fix paket install and update.

### What does this PR do?

 * Changes versions of resources in `index.json` to 3.0.0.
 * Adds trailing `/` in `PackageBaseAddress` and `RegistrationsBaseUrl` which were causing issues in paket too.

### Motivation

Currently running paket install or update would fail with
```
GetPackageDetails failed: System.Exception: Couldn't get package details for package baget-test1 1.0.0 on http://baget:9090/v3/index.json. ---> System.AggregateException: One or more errors occurred. ---> System.Exception: could not find an PackageIndex endpoint for http://baget:9090/v3/index.json
```
Which is caused by resource names in `index.json` being `RegistrationsBaseUrl/Versioned` rather than something like `RegistrationsBaseUrl/3.4.0`.

This is not passing all e2e tests yet.

### DIFF
--- a/src/BaGet.Web/Controllers/IndexController.cs
+++ b/src/BaGet.Web/Controllers/IndexController.cs
@@ -1,4 +1,6 @@
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using BaGet.Web.Extensions;
 using Microsoft.AspNetCore.Mvc;
 using Newtonsoft.Json;
@@ -10,31 +12,31 @@ namespace BaGet.Web.Controllers
     /// </summary>
     public class IndexController : Controller
     {
+        private IEnumerable<ServiceResource> ServiceWithAliases(string name, string url, params string[] versions) 
+        {
+            foreach (var version in versions) 
+            {
+                string fullname = string.IsNullOrEmpty(version) ? name : name + "/" + version;
+                yield return new ServiceResource(fullname, url);
+            }
+        }
+
         // GET v3/index
         [HttpGet]
         public object Get()
         {
             // Documentation: https://docs.microsoft.com/en-us/nuget/api/overview
-            // NuGet.org: https://api.nuget.org/v3-index/index.json
-
-            // TODO: Client is picky about versions' formatting. Needs more research. See:
-            // https://github.com/NuGet/NuGet.Client/blob/67d34ee8158597377fb26e06987ad6ad40a1b09b/src/NuGet.Core/NuGet.Protocol/Constants.cs
+            // NuGet.org: https://api.nuget.org/v3-index/index.json            
             return new
             {
-                Version = "3.0.0-beta.1",
-                Resources = new[]
-                {
-                    // Required
-                    new ServiceResource("PackagePublish/Versioned", Url.PackagePublish()),
-                    new ServiceResource("SearchQueryService/Versioned", Url.PackageSearch()),
-                    new ServiceResource("RegistrationsBaseUrl/Versioned", Url.RegistrationsBase()),
-                    new ServiceResource("PackageBaseAddress/Versioned", Url.PackageBase()),
-
-                    // Optional
-                    new ServiceResource("SearchAutocompleteService/Versioned", Url.PackageAutocomplete()),
-                    //new ServiceResource("ReportAbuseUriTemplate/3.0.0-rc", new Uri("https://google.com")),
-                    //new ServiceResource("Catalog/3.0.0", new Uri("https://google.com"))
-                }
+                Version = "3.0.0",
+                Resources = 
+                    ServiceWithAliases("PackagePublish", Url.PackagePublish(), "2.0.0") // api.nuget.org returns this too.
+                    .Concat(ServiceWithAliases("SearchQueryService", Url.PackageSearch(), "", "3.0.0-beta", "3.0.0-rc")) // each version is an alias of others
+                    .Concat(ServiceWithAliases("RegistrationsBaseUrl", Url.RegistrationsBase(), "", "3.0.0-rc", "3.0.0-beta"))
+                    .Concat(ServiceWithAliases("PackageBaseAddress", Url.PackageBase(), "3.0.0"))
+                    .Concat(ServiceWithAliases("SearchAutocompleteService", Url.PackageAutocomplete(), "", "3.0.0-rc", "3.0.0-beta"))
+                    .ToList()
             };
         }
 
--- a/src/BaGet.Web/Extensions/UrlExtensions.cs
+++ b/src/BaGet.Web/Extensions/UrlExtensions.cs
@@ -6,8 +6,8 @@ namespace BaGet.Web.Extensions
 {
     public static class UrlExtensions
     {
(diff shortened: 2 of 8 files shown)


src/BaGet.Web/Controllers/IndexController.cs
```diff
@@ -19,19 +19,20 @@ public object Get()
...
-                Version = "3.0.0-beta.1",
+                Version = version,
                 Resources = new[]
                 {
                     // Required
                     new ServiceResource("PackagePublish/Versioned", Url.PackagePublish()),
-                    new ServiceResource("SearchQueryService/Versioned", Url.PackageSearch()),
-                    new ServiceResource("RegistrationsBaseUrl/Versioned", Url.RegistrationsBase()),
-                    new ServiceResource("PackageBaseAddress/Versioned", Url.PackageBase()),
+                    new ServiceResource("SearchQueryService/" + version, Url.PackageSearch()),
```
loic-sharma: This breaks Visual Studio's search as `SearchQueryService/3.0.0` isn't a valid resource (the valid resources according to [the documentation](https://docs.microsoft.com/en-us/nuget/api/search-query-service-resource) are `SearchQueryService`, `SearchQueryService/3.0.0-beta`, `SearchQueryService/3.0.0-rc`).

loic-sharma:
Thanks for testing BaGet with Paket! I think that we should have an entry for all documented versions of a resource, otherwise, we'll have a hard time supporting all versions of NuGet/Visual Studio/.NET CLI/Paket.

This solution isn't strictly correct for the package metadata resource as v3.4.0 and v3.6.0 actually have different requirements (see [this documentation](https://docs.microsoft.com/en-us/nuget/api/registration-base-url-resource#registrationsbaseurl340)). I think we can tackle that later when necessary.

Does that sound reasonable to you?

tomzo:
> we should have an entry for all documented versions of a resource,

OK, so many entries to the same resource. I'll compare it with what api.nuget returns too.

> This isn't strictly correct for the package metadata resource as different versions have different requirements

Would you say that current endpoint is compliant with [RegistrationsBaseUrl, RegistrationsBaseUrl/3.0.0-beta, RegistrationsBaseUrl/3.0.0-rc](https://docs.microsoft.com/en-us/nuget/api/registration-base-url-resource#registrationsbaseurl) ?

tomzo:
OK this is now ready. I did the research on accepted versions and this is the best I could implement.
It actually works with `paket` and `dotnet nuget` cli and `mono nuget.exe`. It is very close to what nuget.org returns. I just skipped 3.4.0 and 3.6.0 for `RegistrationsBaseUrl`.

I have also added some integration tests which use client from `NuGet.Protocol` to explore the index.

I had to include change by @WernerMairl  to fix issues with index controller being not used in tests.

loic-sharma:
Thanks! I'll review this when I'm back home.

I asked the team about the `/Versioned` service index types, and was pointed to [this closed documentation PR](https://github.com/NuGet/docs.microsoft.com-nuget/pull/548):

> When the `@type` is suffixed with the `/Versioned` string, an additional property is required on the resource: `clientVersion`.
>
> This value is the minimum client version compatible with this resource. ... A client implementation knows its own version ... and compares it to the resource `clientVersion` value using standard SemVer 2.0.0 comparison rules. If the client's own version is greater than or equal to the resource `clientVersion` and if the `{RESOURCE_NAME}` is known, this resource should be considered consumable by the client implementation.
> 
> For this `clientVersion` model to work, there needs to be some agreement between the server and client implementors about what protocol each `{RESOURCE_NAME}` and `clientVersion` combination implies.
> 
> This feature allows server implementations to backport resources to clients that have already shipped. When no `clientVersion` is specified on the resource, there is no way for a client that is already shipped to discover the resource since the `@type` values that client knows about are hard coded.

src/BaGet.Web/Controllers/IndexController.cs
```diff
@@ -10,31 +12,33 @@ namespace BaGet.Web.Controllers
     /// </summary>
     public class IndexController : Controller
     {
+        private IEnumerable<ServiceResource> ServiceWithAliases(string name, string url, params string[] versions) {
```
loic-sharma: Let's stick to .NET Foundation's style guide (see [this](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md)):

1. `{` on new lines
2. Spaces between `foreach`/`if` and `(`

src/BaGet.Web/Controllers/IndexController.cs
```diff
@@ -10,31 +12,33 @@ namespace BaGet.Web.Controllers
...
-                    new ServiceResource("SearchQueryService/Versioned", Url.PackageSearch()),
-                    new ServiceResource("RegistrationsBaseUrl/Versioned", Url.RegistrationsBase()),
-                    new ServiceResource("PackageBaseAddress/Versioned", Url.PackageBase()),
-
-                    // Optional
-                    new ServiceResource("SearchAutocompleteService/Versioned", Url.PackageAutocomplete()),
-                    //new ServiceResource("ReportAbuseUriTemplate/3.0.0-rc", new Uri("https://google.com")),
-                    //new ServiceResource("Catalog/3.0.0", new Uri("https://google.com"))
-                }
+                Version = version,
```
loic-sharma: The `version` variable seems unnecessary now. How about just `Version` = "3.0.0",`?

src/BaGet.Web/Controllers/IndexController.cs
```diff
@@ -10,31 +12,33 @@ namespace BaGet.Web.Controllers
...
+                Resources = 
+                    ServiceWithAliases("PackagePublish", Url.PackagePublish(), 
+                        "2.0.0", // api.nuget.org returns this too.
+                        "Versioned") // in practice some clients understand this.
```
loic-sharma: `"2.0.0"` should be enough for `PackagePublish`.

loic-sharma:
The [CI build](https://travis-ci.org/loic-sharma/BaGet/builds/436217662?utm_source=github_status&utm_medium=notification) failed, it looks like the [`Mirror` configuration](https://github.com/loic-sharma/BaGet/blob/master/src/BaGet/appsettings.json#L18) is missing from the test server. You should be able to fix that by adding a config [here](https://github.com/loic-sharma/BaGet/blob/master/tests/BaGet.Tests/TestServerBuilder.cs#L85):

```
private readonly string MirrorEnabledKey = $"{nameof(BaGetOptions.Mirror)}:{nameof(MirrorOptions.Enabled)}";

...

Configuration.Add(MirrorEnabledKey , false.ToString());
```

The change looks good though! :)

tomzo:
> The CI build failed, it looks like the Mirror configuration is missing from the test server. You should be able to fix that by adding a config here:

That's also fixed in #81 , so I cherry-picked another commit.

tomzo:
OK, I fixes the style issues and removed both `/Versioned` endpoints ( in push and baseurl resources ).

loic-sharma:
Thanks @tomzo :)

PR #101, opened 2018-09-27
Improved development environment

solidsloth:
### What does this PR do?

In development the application can now be run to proxy to the react development server. This resolves the problem where the code needs to be changed in development to point to the server correcly (which is running on a different port). I also modified the csproj file to have msbuild copy the required front-end files to the publish folder when needed. Therefore the built files no longer need to be copied over and permanently stored in the wwwroot because in development the requests are proxied, and in production the files are copied to the publish folder.

### Closes Issue(s)

I think [EnsureCreated](https://github.com/loic-sharma/BaGet/compare/master...solidsloth:improve-develop#diff-9ad6d4f8918b56df444e175cefd23768R47) this might help with issue #96 because the database is now created in development. 

### Motivation

My main motiviation was to simplify the development and resole the issue about [bend URLs being broken in develop mode](https://github.com/loic-sharma/BaGet/commit/1466885120f4c19621636ecdf8019145abf53ea6#diff-d6ae89e5d09a8b55b7470a25d69deb31R20). This allows the same workflow: `yarn develop` + `dotnet run` without needing to change any urls in `components/SearchResults.tsx`  or `components/DisplayPackage.tsx`

### Additional Notes

- I fixed the vscode `launch.json` file to work with the new netcoreapp2.1.
- I added a few artifacts to the gitignore to prevent accidental commits of the generated files, including: the .cache folder, the dist folder, and the SQLite database file.
- I removed the generated source files from the repo since these should be rebuilt upon every release.

PR #105, opened 2018-10-01
Docker the right way

tomzo:
This PR is on top of CD setup in #103. 

### What does this PR do?

 * It changes the way to build docker image. By always using the result of `msbuild /t:Publish` in the docker image. And we are not using docker BUILD container, but using the same artifacts that were tested in previous stage.
 * added a flag to skip running db migrations. By default it is ran. As mentioned in discussion in #101
 * added e2e tests, mostly ported from LiGet

Dockerfile is changed a bit:
 * uses tini for init system to reap any processes.
 * does not run as root by default. There is a `baget` user.
 * because of above changes default listening port to 9090.
 * when container starts it changes uid and gid of `baget` user based on owner of volume mount.
 * environment variables are all specified in Dockefile for clarity. Now a basic server to try out can be run with `docker run baget`.

### Motivation

The e2e tests ensure that some of the most basic commands, such as nuget push or paket install pass.
This step would be great to execute before publishing final docker image.
Most of this is LiGet's CD ported to BaGet.
The files in `.build` are part of our internal bash script helpers used for common stuff. This is added for reference, nothing fancy there.


tomzo:
Opened #108 instead

Issue #57, opened 2018-07-12
Add Continuous Delivery


Baklap4:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?
Currently there's no continuous delivery. In other words this means we have to manually deploy our build outputs produced in #56. We don't want to do manual stuff as we're programmers so why not automate it? 

### Blocked by
* Blocked by #56 

### Expectations
The CD should be able to do:
- [ ] Deploy build outputs
    - [ ] Windows
    - [ ] Linux
    - [ ] OSx
- [ ] Create a new tag on GitHub
- [ ] Run nightly deploy

This will gives us a place to download artifacts from per version and a place to download the nightly build from.

smazurov:
I think for this, publishing a docker image (under latest/nightly tags) would satisfy the requirements.

tomzo:
I have implemented a pipeline which runs all tests and publishes in docker image only when all have passed. You can see it working in [my fork](https://github.com/ai-traders/baget).
Most of work for it is in PR #108

loic-sharma:
I'd like to follow NuGet Package Explorer's setup:

1. Use [Nerdbank.GitVersioning](https://github.com/aarnott/Nerdbank.GitVersioning) to generate a version
2. Publish BaGet as Azure DevOps artifacts
3. With human approval, use [github-tools-vsts](https://github.com/marceloavf/github-tools-vsts)'s `Create or Modify GitHub Release` task to release the artifacts to GitHub
4. With human approval, publish a Docker image

loic-sharma:
I've added Continuous Delivery to BaGet. You can:

* Download BaGet as a GitHub release: https://github.com/loic-sharma/BaGet/releases
* Install BaGet's components using NuGet: [`BaGet.Core`](https://www.nuget.org/packages/BaGet.Core) and [`BaGet.Azure`](https://www.nuget.org/packages/BaGet.Azure/0.1.1-prerelease)

I'll add support for Docker images in the future.

BlythMeister:
![](https://media.giphy.com/media/5GoVLqeAOo6PK/200.gif)

loic-sharma:
Closing this issue as BaGet has continuous delivery. Please let me know if you have any other suggestions!

Issue #56, opened 2018-07-12
Add Continuous Integration


Baklap4:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?
Currently there's no Continuous integration. In other words this means we have to run the unit tests ourselves, check style coding etc, check if it builds. I bet we don't want that. We'd love to see this repetitive stuff to be automated as much as possible. Here comes CI around the corner.

### Expectations
The CI should be able to do:
- [ ] Be triggered upon a push (commit or tags)
- [ ] Restore dependency packages
    - Run the command `dotnet restore --packages SOME/LOCATION` in the same folder as where the `sln file` is.
- [ ] Cache the dependencies so our buildtime will decrease
    - Run the previous command with the cached packages when the pipeline builds a second time.
- [ ] Build the project
    - [ ] Windows
    - [ ] Linux
    - [ ] OSx
- [ ] Run tests
    - [ ] Windows
    - [ ] Linux
    - [ ] OSx
- [ ] Once every night run a nightly build.

If the CI meets these expectations AND the push is a tag/release CD will pop up to release a new version. see #57

Baklap4:
@loic-sharma i'd love to set this up, however i'd need access to the webhook area of this github page

Baklap4:
Hmm personally i'm not a big fan of VSTS, since it was so hard to configure and let it do what it should be doing. Let be creating a docker image or such.. therefore I moved to gitlab which was way easier. Here on Github most of the people do use travis or jenkins to do what they want

WernerMairl:
I have my own CI Build System here (Jenkins, Docker, implemented in pwsh on Linux)
 
I like the idea to define/implement a few basic entry points and scripts that can be used from multiple 
CI-Systems.....

to avoid deep/hard dependencies into ONE CI-Build system, so we are open for

- VSTS (Pipeline or Classic)
- travis et. al
- custom/on premise systems

that may by a litle bit more of work, but it should bring back a lot of learnings in this area!

what you think about that ?

loic-sharma:
Added Travis CI as part of https://github.com/loic-sharma/BaGet/commit/a4ddace43eb4ff2cca19bcd411a809e23813d358. Feel free to open another issue if you have any other ideas!

loic-sharma:
Reopening this. The current set up doesn't run unit tests.

loic-sharma:
Travis will now run the tests as part of the build. Closing this again. Apologies for the noise!

[...]

PR #108, opened 2018-10-04
Improve build system; add e2e tests, docker CD, FAKE, paket

tomzo:
### What does this PR do?

 * This PR contains a rebase and some fixes of #101 . @solidsloth I hope you are fine with me rebasing your commit and integrating with FAKE. In #101 quite a few things did not want to work at the same time. You could either get debugging work, tests or production build. I have taken care of this now. 
 * Added test which checks that built image returns contents of SPA site
 * It includes all of #103, which is:
   1. Adds `Idefile`, which is a reference to [docker IDE image](https://github.com/ai-traders/docker-dotnet-ide), which contains all build dependencies (dotnet, node.js, yarn, bats for e2e tests). If you have local docker daemon then you can run any of the build tasks like this now `ide <command>`. E.g. `ide ./build.sh` will run msbuild publish and run tests with xunit.
   2. Updates travis setup to run:
       - msbuild publish in release config
       - run tests of the published artifacts
       - build docker image from those artifacts
       - test that docker image is functional
   2. Moves dependency management to paket.
   3. Adds FAKE and targets to run most of above workflow.
 * It includes all of #105, which is:
    * It changes the way to build docker image. By always using the result of `msbuild /t:Publish` in the docker image. And we are not using docker BUILD container, but using the same artifacts that were tested in previous stage.
    * added a flag to skip running db migrations. By default it is ran. As mentioned in discussion in #101
    * added e2e tests, mostly ported from LiGet. 
       - We create some packages, push, download with nuget cli and paket. 
       - These can run on travis from now on

Dockerfile is changed a bit:
 * uses tini for init system to reap any processes.
 * does not run as root by default. There is a `baget` user.
 * because of above changes default listening port to 9090.
 * when container starts it changes uid and gid of `baget` user based on owner of volume mount.
 * environment variables are all specified in Dockefile for clarity. Now a basic server to try out can be run with `docker run baget`.

### Closes Issue(s)

#57 #56 #101 #7 

### Motivation

In general - I need BaGet to be functional and tested before I can use it in production. I don't want to waste time on testing features and fixing bugs which get randomly broken because there are no tests, no reproducible builds and no release cycle.

Why `Idefile` and [IDE project](https://github.com/ai-traders/ide):
 - biggest advantage is to reduce the noise around works on my machine. Now anyone can download the same environment which all developers and travis agents are running for builds.
 - rather than spending time on setting up your host, (either laptop or the CI agent) just pull the image which someone has already created and tested to be sufficient to execute entire workflow of the project

Why paket:
 - paket gets consistent package versions across all projects and it can support modules for FAKE.
 - `paket.dependencies` gives you an quick overview of all packages used by application. 
- `paket.lock` gives you detailed info about each version of all direct and transitive dependencies.
 - I tried to install xunit.runner.console without using paket. But there is no such command on `dotnet` SDK. There is no nuget CLI for linux which works without mono, and I managed to keep away from mono so far.

Why FAKE:
 - we get consistent AssemblyVersion for all projects, from top of changelog file.
 - it provides single source of truth on what the release cycle and tasks involved.
 - fake is cross platform and can do much more than just msbuild. (See how hard it was to copy a few files in #101)
 - programming tasks in xml is not easy and error prone. In F# FAKE you get as much as type safety in build scripts.
 - for all other task, which cannot be covered by msbuild or dotnet cli, we need a cross platform language. Otherwise windows guys start to code in powershell and linux guys in bash. We need something friendly to both platforms.
 - F# is cool

Why I am running tests with `dotnet ...xunit.console.dll <assemblies-from-publish-dir>`?
The main point in CD is that artifacts which constitute the application should be produced just once and then ran through all QA stages. We really need to be testing exactly the same files at each point until final release to ensure quality. In dotnet world that means using `dotnet publish` to force outputting application and all all its dependencies.
One of the common issues I had with dotnet assemlies is that running with dotnet test adapter uses additional assemblies from nuget caches. So I've seen tests passing when ran with `dotnet test` or `dotnet xunit`, but when same test is executed with  `dotnet ...xunit.console.dll <assemblies-from-publish-dir>` it fails. Naturally same load errors happen in deployed application (after all tests have passed earlier...). So running tests with `publish` reproduces those errors and makes sure that we get load errors early.
PS: I actually ran into this already in BaGet and fixed it. The `Microsoft.AspNetCore.App` in BaGet.Tests was causing problems like this:
```
xUnit.net Console Runner v2.4.0 (64-bit .NET Core 4.6.26814.03)
System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.Extensions.Logging.Abstractions, Version=2.1.1.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. The system cannot find the file specified.

System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.Extensions.Logging.Abstractions, Version=2.1.1.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. The system cannot find the file specified.
```

### Additional Notes

I will be away for weekend, so I can have a look at review notes on Monday.
I am closing #103 and #105 , and I hope @solidsloth can close #101 since this contains the work and original intentions.


loic-sharma:
Hey @tomzo, I apologize for the incredibly long response time. I feel conflicted with these changes as I want to keep BaGet as simple as possible.

**IDE project:**
While I see the value in a reproducible development environment, I don't feel comfortable taking a dependency on this project. I'm not particularly familiar with IDE, and I'd rather focus my time on the features that differentiate BaGet. I don't think that setting up a development environment to work on BaGet is too difficult, and I'd like to keep it that way.

**Paket:**
Paket is a great project and BaGet should work well with Paket's ecosystem. However, I work on NuGet and would like to stick to NuGet 😄 

> paket gets consistent package versions across all projects 
> `paket.dependencies` gives you an quick overview of all packages used by application.

This is possible using NuGet too (see [the aspnet repository](https://github.com/aspnet/AspNetCore/blob/master/build/dependencies.props) for an example). NuGet will be improving this experience in the future: [Centrally managing NuGet package versions](https://github.com/NuGet/Home/wiki/Centrally-managing-NuGet-package-versions). The `paket.dependencies` file is indeed pretty neat!

> `paket.lock` gives you detailed info about each version of all direct and transitive dependencies.

Yes, this a feature that NuGet is sorely lacking. We are actively working on this, see [this announcement](https://github.com/NuGet/Home/wiki/Enable-repeatable-package-restore-using-lock-file).

> There is no nuget CLI for linux which works without mono, and I managed to keep away from mono so far.

Out of curiosity, why do you want to use the nuget CLI on linux? The dotnet CLI should have most everything you need.

**FAKE:**

> fake ... can do much more than just msbuild. (See how hard it was to copy a few files in #101)
> programming tasks in xml is not easy and error prone. In F# FAKE you get as much as type safety in build scripts.

I 110% agree. I'm very torn about this as I very much dislike MSBuild. However, I want to keep BaGet as approachable as possible to all .NET developers. I don't think FAKE fits well with this goal.

> for all other task, which cannot be covered by msbuild or dotnet cli, we need a cross platform language. Otherwise windows guys start to code in powershell and linux guys in bash. We need something friendly to both platforms.

I agree that we should avoid this. If possible,  I'd like to keep all tasks in MSBuild (😢). Is there anything in particular that you think would be too crazy for MSBuild?

Okay, I'm gonna stop here as I'd love to hear your thoughts @tomzo.

tomzo:
> Paket ... This is possible using NuGet too

Without `paket.lock` the build is not reproducible, because of transitive dependencies not being locked. Which is not acceptable for me, for any project which I intend to deploy. Maybe NuGet will be capable of this one day, that's great. But a solution now is paket.

> Out of curiosity, why do you want to use the nuget CLI on linux? The dotnet CLI should have most everything you need.

nuget cli has `install` command which works as downloader of nupkg. Other than I don't need it.
I had to download xunit console runner, to run tests as I explained earlier. This helps to reproduce assembly load errors such as https://github.com/loic-sharma/BaGet/issues/112 during tests.

> I agree that we should avoid this. If possible, I'd like to keep all tasks in MSBuild (cry). Is there anything in particular that you think would be too crazy for MSBuild?

I don't see a problem in running msbuild for restore, build and test. Because those tasks are trivial. 
Have a look all tasks I implemented in https://github.com/ai-traders/BaGet/blob/master/build.fsx for managing release cycle, changelog, git tags, publishing to github. Do you really intend to have those in msbuild?

> Okay, I'm gonna stop here as I'd love to hear your thoughts @tomzo. I realize that I haven't addressed your test and Docker changes. I'm much more onboard with these changes, so that should be smoother :)

But you are saying that we cannot use FAKE, paket and IDE. Each of those is critical to make those end tests work. Sorry, but I won't be rebasing those tests without any framework, because it would be a lot of wasted time on setups, which would only poorly mimic the tools needed to get the job done.

My initial intent was to replace LiGet by merging most of it into BaGet and then move to it. I have mostly done it already in my fork. But now it looks to me that we have different objectives about nuget server project.
You want to keep it as simple as possible and easy for (I guess) Visual Studio developers. I think that you are also OK with this being still an experimental, PoC project. I need a working, maintainable product, which BaGet is not, because it is not tested and released.
So in my opinion, the best to do here is split ways again. I already have 1/2 of LiGet in BaGet fork, so I will just port it back to LiGet and keep my build system, tests and releases, which I had from LiGet anyway.
I think we agree on adding more tests coverage, so please feel free to cherry-pick some of tests which I added from my fork - https://github.com/ai-traders/BaGet/tree/master/tests/BaGet.Tests

BlythMeister:
As someone external, I hope that we can find some middle ground on this.
Whilst I am a huge advocate of paket and fake (and a contributor to both) I can see why there would be apprehension to add to this project.

The 2 things I think this project is missing is the end to end tests to prove it works the same as nuget.

The ability for someone to pick up a docker image, or package to host themselves.
For me, I don't want to host I docker, but spent around an hour trying to get something which I could self host.

My suggestion is to work out how to add these tests and other changes which add value to the project, and then investigate other build tools for example appveyor or Azure pipelines.
Which both support the ability to create things like GitHub releases without the need for fake.

loic-sharma:
>  I need a working, maintainable product, which BaGet is not, because it is not tested and released.
So in my opinion, the best to do here is split ways again.

That's completely fair. I'd like to investigate @BlythMeister's suggestion of Azure pipelines for releases.

> I think we agree on adding more tests coverage, so please feel free to cherry-pick some of tests which I added from my fork - https://github.com/ai-traders/BaGet/tree/master/tests/BaGet.Tests

Will do . Thank you for the great work.

> The ability for someone to pick up a docker image, or package to host themselves.
For me, I don't want to host I docker, but spent around an hour trying to get something which I could self host.

This sounds interesting @BlythMeister . Do you know of other projects that support self hosting through packages? I'd love to mimic their setup.

loic-sharma:
I've added Continuous Delivery to BaGet. You can:

* Download BaGet as a GitHub release: https://github.com/loic-sharma/BaGet/releases
* Install BaGet's components using NuGet: [`BaGet.Core`](https://www.nuget.org/packages/BaGet.Core) and [`BaGet.Azure`](https://www.nuget.org/packages/BaGet.Azure/0.1.1-prerelease)

I'll add support for Docker images in the future. Please let me know what you think!

PR #114, opened 2018-11-02
Add configuration option for application root URL

daniel-lerch:
### What does this PR do?

This pull request adds the setting `PathBase` to the BaGet configuration and applies this setting on startup. The `PathBase` option allows an ASP.NET Core application to run under a different root URL than it was designed for.

### Motivation

I want to operate a BaGet server with a URL like https://example.com/baget and reach the API index at https://example.com/baget/v3/index.json. With this change I don't have to write complex rewrite rules for this scenario which might not work at all.

### Additional Notes

[Deal with path base and proxies that change the request path](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-2.1#deal-with-path-base-and-proxies-that-change-the-request-path) on Microsoft Docs


src/BaGet.Core/Configuration/BaGetOptions.cs
```diff
@@ -8,6 +8,11 @@ public class BaGetOptions
         /// </summary>
         public string ApiKeyHash { get; set; }
 
+        /// <summary>
+        /// The application root URL for usage in reverse proxy scenarios.
+        /// </summary>
+        public string PathBase { get; set; }
```
loic-sharma: Let's give the `PathBase` option a default value of `""` so that we can remove the config from `appsettings.json`. I'd like to keep the `appsettings.json` file uncluttered.

src/BaGet/Startup.cs
```diff
@@ -46,6 +47,7 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
                 }
             }
 
+            app.UsePathBase(Configuration.Get<BaGetOptions>().PathBase);
```
loic-sharma: Does this throw if `PathBase` is `null`? If it does, please add a null check.

src/BaGet/appsettings.json
```diff
@@ -1,5 +1,6 @@
 ﻿{
   "ApiKeyHash": "",
+  "PathBase": "",
```
loic-sharma: Let's remove the `PathBase` option out of the default `appsettings.json` file. I'd like to keep the default `appsettings.json` file as simple as possible as to not overwhelm new users.

loic-sharma:
Looks great! I left some minor comments, I'll merge once you've addressed them :)

src/BaGet/Startup.cs
daniel-lerch: This option does not throw on `null`.

daniel-lerch:
This PR will take longer than expected. I experienced problems with the configuration as some URLs are kind of hard-coded and are not adjusted by changing the `PathBase`. I'll write another comment when this is fixed.

daniel-lerch:
The API can now be configured to run under a different `PathBase`. I only had to change the manual URLs generation.

For the web UI it is not that easy to configure the base path as the site is served as static content. Two ways to resolve this problem:
1. We could change the links to `Baget.UI.xyz` instead of `/Baget.UI.xyz`
2. We create a new controller for the web UI which replaces these variables and serves the files. With a memory cache we won't loose performance but I don't want to implement this major change myself.

loic-sharma:
Ah I see. Today, the frontend's API endpoints are hardcoded in the code base. I'd like to make the backend's endpoint configurable at build time which should unblock you. My recommendation for you:

1. Make a temporary fork of BaGet
2. Update the URIs in the frontend:
  a. [DisplayPackage.tsx#L55](https://github.com/loic-sharma/BaGet/blob/master/src/BaGet.UI/components/DisplayPackage.tsx#L55)
  b. [DisplayPackage.tsx#L112](https://github.com/loic-sharma/BaGet/blob/master/src/BaGet.UI/components/DisplayPackage.tsx#L112)
  c. [SearchResults.tsx#L71](https://github.com/loic-sharma/BaGet/blob/master/src/BaGet.UI/components/SearchResults.tsx#L71)
  d. [SearchResults.tsx#L72](https://github.com/loic-sharma/BaGet/blob/master/src/BaGet.UI/components/SearchResults.tsx#L72)
3. Build the frontend using [these instructions](https://github.com/loic-sharma/BaGet/blob/master/src/BaGet.UI/readme.md)

This is a short-term solution that should let you host BaGet in a different root URL. I'll work on improving the frontend's setup soon.

daniel-lerch:
Is there anything missing to merge this PR or do you want to wait until the frontend will work with a different root URL?

loic-sharma:
Your changes look good. I moved some things around after you opened this pull request - would you mind applying your changes on the latest master? I'll go ahead and merge after :)

daniel-lerch:
I just merged your changes into my fork.

loic-sharma:
Thank you :)

cinderell4:
Hello,

I'm trying to use this, but it is not working right now.

appsettings.json:
`
{
[...]
  "PathBase": "/nuget",
[...]
}
`
Can u help me to understand, how I've to set this option?
Thx in advance.

daniel-lerch:
The `PathBase` option only affects the API which is served by ASP.NET Core. You should be able to access [http://localhost:5000<b>/nuget</b>/v3/index.json](http://localhost:5000/nuget/v3/index.json). Visual Studio should work with this path base as expected.

However, the Web UI does not work [as mentioned above](#issuecomment-436226881).

cinderell4:
Is there any solution for having BaGet incl. UI behind a proxy? I got the feeling I'm missing something obvious here.
Thx for ur help.

daniel-lerch:
There is currently no way to run BaGet's web UI with a different root URL.

@loic-sharma recommend to change the hard-coded URLs in one's own fork ([see above](#issuecomment-437054902)).

loic-sharma:
I haven’t played around with this, but you should be able to add the `homepage` setting to the [package.json](https://github.com/loic-sharma/BaGet/blob/master/src/BaGet.UI/package.json) file to support your need. See: https://facebook.github.io/create-react-app/docs/deployment

cinderell4:
After trying different things imcluding homepage, basename etc. I've to say, that i'm not making any progress. I need to take a look at this topic again. Thank you for your effort and your help so far.

Issue #119, opened 2018-11-07
Docker build on Windows container fail


bryjamus:
### Steps to reproduce

1. Setup docker to use **windows containers**
2. Build container using:
```
docker build . -t baget
```
You get this error:

```
Step 9/16 : RUN dotnet build src/BaGet -c ${APP_BUILD} -o /app
 ---> Running in cbaa56a45bf7
Microsoft (R) Build Engine version 15.8.169+g1ccb72aefa for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

  Restore completed in 1.99 sec for C:\src\src\BaGet.Core\BaGet.Core.csproj.
  Restore completed in 2.02 sec for C:\src\src\BaGet.Azure\BaGet.Azure.csproj.
  Restore completed in 4.86 sec for C:\src\src\BaGet\BaGet.csproj.
C:\Program Files\dotnet\sdk\2.1.403\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.GenerateAssemblyInfo.targets(134,5): error MSB4018: The "WriteCodeFragment" task failed unexpectedly. [C:\src\src\BaGet.Core\BaGet.Core.csproj]
C:\Program Files\dotnet\sdk\2.1.403\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.GenerateAssemblyInfo.targets(134,5): error MSB4018: System.FormatException: Input string was not in a correct format. [C:\src\src\BaGet.Core\BaGet.Core.csproj]
C:\Program Files\dotnet\sdk\2.1.403\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.GenerateAssemblyInfo.targets(134,5): error MSB4018:    at System.Text.StringBuilder.FormatError() [C:\src\src\BaGet.Core\BaGet.Core.csproj]
C:\Program Files\dotnet\sdk\2.1.403\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.GenerateAssemblyInfo.targets(134,5): error MSB4018:    at System.Text.StringBuilder.AppendFormatHelper(IFormatProvider provider, String format, ParamsArray args) [C:\src\src\BaGet.Core\BaGet.Core.csproj]
[...]

PR #120, opened 2018-11-08
Use Release configuration in docker images

loic-sharma:
Fixes https://github.com/loic-sharma/BaGet/issues/119

/cc @bryjamus


bryjamus:
Thx

Issue #130, opened 2018-11-16
Make BaGet a symbol server


loic-sharma:
BaGet should:

1. Have an endpoint to ingest [symbol packages (.snupkg)](https://docs.microsoft.com/en-Us/nuget/create-packages/symbol-packages-snupkg)
2. It should be able to extract portable PDBs embedded into a nupkg
3. It should be able to respond to symbol server requests. See [this API documentation](https://docs.microsoft.com/en-us/rest/api/azure/devops/symbol/?view=azure-devops-rest-4.1)

See https://github.com/dotnet/symstore/blob/master/docs/specs/Simple_Symbol_Query_Protocol.md
See https://github.com/dotnet/symstore/blob/master/docs/specs/SSQP_Key_Conventions.md
See https://github.com/dotnet/corefx/blob/master/src/System.Reflection.Metadata/specs/PE-COFF.md
See https://github.com/dotnet/corefx/blob/master/src/System.Reflection.Metadata/specs/PortablePdb-Metadata.md

See https://github.com/dotnet/symstore/tree/master/src/NugetSymbolServer

/cc @karann-msft

SwooshyCueb:
Duplicate of #11?

loic-sharma:
Oops, good catch! This issue includes snupkg ingestion which was just released yesterday in Visual Studio 15.9, so I’ll close the other issue.

loic-sharma:
Finished by https://github.com/loic-sharma/BaGet/pull/134

PR #134, opened 2018-11-20
Make BaGet a symbol server

loic-sharma:
Allows BaGet to: 

* Index symbols from [`.snupkg`](https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg) packages
* Serve symbols using the [Simple Symbol Query Protocol](https://github.com/dotnet/symstore/blob/master/docs/specs/Simple_Symbol_Query_Protocol.md#simple-symbol-query-protocol-ssqp)

See: https://github.com/loic-sharma/BaGet/issues/130


src/BaGet/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -147,6 +148,30 @@ public static IServiceCollection ConfigureHttpServices(this IServiceCollection s
...
+
+                storageOptions.EnsureValid();
+
+                switch (storageOptions.Type)
+                {
+                    case StorageType.FileSystem:
+                        return provider.GetRequiredService<FileSymbolStorageService>();
+
+                    // TODO
+                    case StorageType.AzureBlobStorage:
```
loic-sharma: TODO

Issue #25, opened 2018-04-16
Redirect for downloads


loic-sharma:
Today, package content downloads consume a thread for the duration of the download. This will not scale well. Instead, download endpoints should redirect to Azure Blob Storage/CDN.

LordMike:
Ideally necessary for #136

PR #144, opened 2018-12-02
Add S3 storage service

### LINKED ISSUE #136: AWS support (S3)
Hi,

What is needed for S3 support? I can see the Azure project has a few things in it:

* Blob service
* Search service using Azure Search

This directly translates to:

* S3 storage (can we do it so that end users download from S3 directly?)
* AWS CloudSearch (never used it)
  * Is search necessary? Wouldn't a query in the EF DB be enough?
  * I imagine most queries want to prefix match on the Package Id anyways

Would implementing an S3 client be enough?

LordMike:
(do not merge)

### What does this PR do?

Implements an S3 storage service

### Closes Issue(s)

Fixes #136

### Motivation

If we deploy BaGet, we'll need S3 storage support :)

### Additional Notes

This code has not been tested, yet. I'd like some feedback before proceeding. It's basically a copy off of the Azure project, with the search removed. I've replicated the configuration validation methods etc.

### DIFF
--- a/src/BaGet.AWS/BaGet.AWS.csproj
+++ b/src/BaGet.AWS/BaGet.AWS.csproj
@@ -1,0 +1,18 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
+
+    <Description>The libraries to host BaGet on AWS.</Description>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="AWSSDK.S3" Version="3.3.30" />
+    <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="2.1.1" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\BaGet.Core\BaGet.Core.csproj" />
+  </ItemGroup>
+
+</Project>
--- a/src/BaGet.AWS/Configuration/S3StorageOptions.cs
+++ b/src/BaGet.AWS/Configuration/S3StorageOptions.cs
@@ -1,0 +1,22 @@
+﻿using System.ComponentModel.DataAnnotations;
+using BaGet.Core.Validation;
+
+namespace BaGet.AWS.Configuration
+{
+    public class S3StorageOptions
+    {
+        [RequiredIf(nameof(KeySecret), null, IsInverted = true)]
+        public string KeyId { get; set; }
+
+        [RequiredIf(nameof(KeyId), null, IsInverted = true)]
+        public string KeySecret { get; set; }
+
+        [Required]
+        public string Region { get; set; }
+
+        [Required]
+        public string Bucket { get; set; }
+
+        public string Prefix { get; set; }
+    }
+}
--- a/src/BaGet.AWS/Configuration/S3StorageOptions.cs
+++ b/src/BaGet.AWS/Configuration/S3StorageOptions.cs
@@ -5,11 +5,11 @@ namespace BaGet.AWS.Configuration
 {
     public class S3StorageOptions
     {
-        [RequiredIf(nameof(KeySecret), null, IsInverted = true)]
-        public string KeyId { get; set; }
+        [RequiredIf(nameof(SecretKey), null, IsInverted = true)]
+        public string AccessKey { get; set; }
 
-        [RequiredIf(nameof(KeyId), null, IsInverted = true)]
-        public string KeySecret { get; set; }
+        [RequiredIf(nameof(AccessKey), null, IsInverted = true)]
+        public string SecretKey { get; set; }
 
         [Required]
         public string Region { get; set; }
--- a/src/BaGet.AWS/Extensions/ServiceCollectionExtensions.cs
+++ b/src/BaGet.AWS/Extensions/ServiceCollectionExtensions.cs
@@ -1,0 +1,35 @@
+﻿using Amazon;
+using Amazon.Runtime;
+using Amazon.S3;
+using BaGet.AWS.Configuration;
+using BaGet.Core.Services;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+
+namespace BaGet.AWS.Extensions
+{
+    public static class ServiceCollectionExtensions
+    {
+        public static IServiceCollection AddS3StorageService(this IServiceCollection services)
+        {
+            services.AddSingleton(provider =>
+            {
+                var options = provider.GetRequiredService<IOptions<S3StorageOptions>>().Value;
+
+                AmazonS3Config config = new AmazonS3Config
+                {
+                    RegionEndpoint = RegionEndpoint.GetBySystemName(options.Region)
+                };
+
+                if (!string.IsNullOrEmpty(options.KeyId))
+                    return new AmazonS3Client(new BasicAWSCredentials(options.KeyId, options.KeySecret), config);
+
+                return new AmazonS3Client(config);
+            });
+
+            services.AddTransient<S3StorageService>();
+
+            return services;
+        }
+    }
+}
--- a/src/BaGet.AWS/Extensions/ServiceCollectionExtensions.cs
+++ b/src/BaGet.AWS/Extensions/ServiceCollectionExtensions.cs
@@ -2,7 +2,6 @@
 using Amazon.Runtime;
 using Amazon.S3;
 using BaGet.AWS.Configuration;
-using BaGet.Core.Services;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
 
@@ -21,8 +20,8 @@ namespace BaGet.AWS.Extensions
                     RegionEndpoint = RegionEndpoint.GetBySystemName(options.Region)
                 };
 
-                if (!string.IsNullOrEmpty(options.KeyId))
-                    return new AmazonS3Client(new BasicAWSCredentials(options.KeyId, options.KeySecret), config);
+                if (!string.IsNullOrEmpty(options.AccessKey))
+                    return new AmazonS3Client(new BasicAWSCredentials(options.AccessKey, options.SecretKey), config);
 
                 return new AmazonS3Client(config);
             });
--- a/src/BaGet.AWS/S3StorageService.cs
+++ b/src/BaGet.AWS/S3StorageService.cs
@@ -1,0 +1,98 @@
+﻿using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Amazon.S3;
+using Amazon.S3.Model;
+using BaGet.AWS.Configuration;
+using BaGet.Core.Services;
+using Microsoft.Extensions.Options;
+
+namespace BaGet.AWS
+{
+    public class S3StorageService : IStorageService
+    {
+        private const string Separator = "/";
+        private readonly string _bucket;
+        private readonly string _prefix;
+        private readonly AmazonS3Client _client;
+
+        public S3StorageService(IOptions<S3StorageOptions> options, AmazonS3Client client)
+        {
+            _bucket = options.Value.Bucket;
+            _prefix = options.Value.Prefix;
+            _client = client;
+
+            if (!string.IsNullOrEmpty(_prefix) && !_prefix.EndsWith(Separator))
+                _prefix += Separator;
+        }
+
+        private string PrepareKey(string path)
+        {
+            return _prefix + path.Replace("\\", Separator);
+        }
+
+        public async Task<Stream> GetAsync(string path, CancellationToken cancellationToken = default)
+        {
+            MemoryStream stream = new MemoryStream();
+
+            try
+            {
+                using (GetObjectResponse res = await _client.GetObjectAsync(_bucket, PrepareKey(path), cancellationToken))
+                    await res.ResponseStream.CopyToAsync(stream);
+
+                stream.Seek(0, SeekOrigin.Begin);
+            }
+            catch (Exception)
+            {
+                stream.Dispose();
+
+                // TODO
+                throw;
+            }
+
+            return stream;
+        }
+
+        public Task<Uri> GetDownloadUriAsync(string path, CancellationToken cancellationToken = default)
+        {
+            string res = _client.GetPreSignedURL(new GetPreSignedUrlRequest
+            {
+                BucketName = _bucket,
+                Key = PrepareKey(path)
+            });
+
+            return Task.FromResult(new Uri(res));
+        }
+
+        public async Task<PutResult> PutAsync(string path, Stream content, string contentType, CancellationToken cancellationToken = default)
+        {
+            // TODO: Uploads should be idempotent. This should fail if and only if the blob
+            // already exists but has different content.
+
+            using (MemoryStream ms = new MemoryStream())
+            {
+                await content.CopyToAsync(ms, 4096, cancellationToken);
+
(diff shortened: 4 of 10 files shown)


LordMike:
@loic-sharma I have a few thoughts and questions.

#### Shouldn't the storage services be simpler than this? 

I would imagine that it would be possible to have a storage service inteface that implements the following:

* Get(string path)
* GetDownloadUrl(string path)
* Put(string path, Stream content, string contentType)
* Delete(string path)

Then a service could be built around this, that exposes the Nuspec/Package/Readme specifics. All storage services would be shielded from new package types (.snupkg), and services could expose new features without breaking implementations.

Adding the GetDownloadUrl() is useful for redirections, see #25. S3 would provide pre-signed urls (or whatever it needs to provie). Azure would do the same.

#### Is it necessary to add _all_ storage implementations in the web project?

It seems odd to me to add all services, and then finally "remap" one service by adding it as the intended interface. Wouldn't it be better to have the wanted service register itself as the intended interface directly?

#### Why put the storage clients in the DI framework?

This is more of a preference thing: Why add the Azure Blobs-specific client to the DI, only to pull it out and give it to one specific implementation? It's not like it can be exchanged (no use of interfaces :P) .. so why not create the Azure Blobs client within `BlobPackageStorageService` (and the `AmazonS3Client` in `S3StorageService` for S3)?

LordMike:
Note: I did register the services in the web project, but didn't commit it because I missed it -- I'm seeing a lot of newline changes which I'm trying to figure out.

src/BaGet.AWS/BaGet.AWS.csproj
```diff
@@ -0,0 +1,19 @@
...
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
+
+    <Description>The libraries to host BaGet on AWS.</Description>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="AWSSDK.S3" Version="3.3.30" />
+    <PackageReference Include="JetBrains.Annotations" Version="2018.2.1" />
```
loic-sharma: What's the JetBrains dependency for?

src/BaGet.AWS/Configuration/S3StorageOptions.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace BaGet.AWS.Configuration
+{
+    public class S3StorageOptions
+    {
+        public string KeyId { get; set; }
+        public string KeySecret { get; set; }
+        public string Region { get; set; }
+        public string Bucket { get; set; }
+        public string Prefix { get; set; }
```
loic-sharma: Out of curiosity, what's the benefit of using a prefix instead of a new bucket? Over on Azure we make new containers (the equivalent of buckets) all the time.

src/BaGet.AWS/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+            services.Configure<S3StorageOptions>(configuration.GetSection(nameof(BaGetOptions.Storage)));
+
+            return services;
+        }
+
+        public static IServiceCollection AddS3StorageService(this IServiceCollection services)
+        {
+            services.AddSingleton(provider =>
+            {
+                S3StorageOptions options = provider.GetRequiredService<IOptions<S3StorageOptions>>().Value;
```
loic-sharma: Please use `var` wherever possible

src/BaGet.AWS/S3StorageService.cs
```diff
@@ -0,0 +1,101 @@
...
+using NuGet.Versioning;
+
+namespace BaGet.AWS
+{
+    public class S3StorageService : IStorageService
+    {
+        private readonly string _bucket;
+        private readonly string _prefix;
+        private readonly AmazonS3Client _client;
+        private const string Seperator = "/";
```
loic-sharma: `Separator`

src/BaGet.AWS/S3StorageService.cs
```diff
@@ -0,0 +1,101 @@
...
+            return _prefix + path.Replace("\\", Seperator);
+        }
+
+        public async Task<Stream> GetAsync(string path, CancellationToken cancellationToken = default)
+        {
+            MemoryStream stream = new MemoryStream();
+
+            try
+            {
+                GetObjectResponse res = await _client.GetObjectAsync(_bucket, PrepareKey(path), cancellationToken);
```
loic-sharma: This response should be disposed

src/BaGet.AWS/S3StorageService.cs
```diff
@@ -0,0 +1,101 @@
...
+
+            return Task.FromResult(new Uri(res));
+        }
+
+        public async Task<PutResult> PutAsync(string path, Stream content, string contentType, CancellationToken cancellationToken = default)
+        {
+            // TODO: Uploads should be idempotent. This should fail if and only if the blob
+            // already exists but has different content.
+
+            using (MemoryStream ms = new MemoryStream())
```
loic-sharma: Do you need to copy `content` to a memory stream? Could you instead pass the `content` stream directly to `PutObjectAsync`?

loic-sharma:
Looks good! I left some minor comments. Please address them and I'll merge these changes in :)

src/BaGet.AWS/BaGet.AWS.csproj
LordMike: One of the configuration extensions throws if the input is null, but Jetbrains Resharper from then on assumes that the input _can_ be null. Using the ContractAnnotation from Resharper allows me to explain to R# that if the input is null, it throws - hence it "can't be null". 

I'll remove it :)

src/BaGet.AWS/Configuration/S3StorageOptions.cs
LordMike: When downloading from the S3 bucket directly, its name dictates the URL. To allow more control of this, I prefer using prefixes - since it allows control of the latter part of the URL. 

I can remove it to be consistent?

loic-sharma: You can keep it in, I can see how it would be useful

src/BaGet.AWS/S3StorageService.cs
LordMike: We can't. Sometimes the `content` is a stream that doesn't support seeking, which is necessary for S3, since it requires the Length of the stream. There is a path through the S3 client code that doesn't require this, but I was unsuccesful in hitting it.

Essentially if you set the ContentLength header on the request object to something, it should use that length if acquiring it from `content` fails - but since we don't know the actual length in a DeflateStream, we can only set it to 0 or something very large. 0 doesn't quite work (uploaded data is actually 0 bytes) -- haven't tried setting it to something large.

src/BaGet.AWS/S3StorageService.cs
LordMike: Consider referencing https://github.com/DotNetAnalyzers/IDisposableAnalyzers.

src/BaGet.AWS/S3StorageService.cs
LordMike: Would it be possible to make an editorconfig ?

src/BaGet.AWS/Configuration/S3StorageOptions.cs
LordMike: Since the annotation (and there was just one) isn't there, I've removed the reference.

src/BaGet.AWS/S3StorageService.cs
LordMike: The only way I know of that approaches a "we don't know the length" is to buffer f.ex. ~25 MB at a time, and then do a Multipart Upload. But you'll still be buffering 25 MB at a time -- undoubtedly better than potentially gigabytes at a time.. :P

LordMike:
@loic-sharma I've just noticed ValidateOptions is never called. Neither is Azure's version.

Where should this be used?

src/BaGet.AWS/S3StorageService.cs
LordMike: AWS actually has a utility that does that - `TransferUtility`. Unfotunately, just like last I looked at it, it still doesn't support unknown lengths.

```
System.NotSupportedException: This operation is not supported.
   at System.IO.Compression.DeflateStream.get_Length()
   at Amazon.S3.Transfer.TransferUtilityUploadRequest.get_ContentLength() in E:\JenkinsWorkspaces\v3-trebuchet-release\AWSDotNetPublic\sdk\src\Services\S3\Custom\_bcl+coreclr+pcl\Transfer\TransferUtilityUploadRequest.cs:line 420
   at Amazon.S3.Transfer.TransferUtility.IsMultipartUpload(TransferUtilityUploadRequest request) in ...
```

LordMike: S3's issue tracker still shows no progress on this one - however, it _is_ possible as I mentioned, it's just a lot of code: fill a buffer till we hit limit, check if there is more data in the source (mark last-chunk), then send it, repeat..

loic-sharma:
> @loic-sharma I've just noticed ValidateOptions is never called. Neither is Azure's version.

Oh whoops, good catch! I'd like to rework how the configuration validation is done. Right now, every consumer of a configuration is expected to call `ValidateOptions`. Maybe the validation should be registered when the configuration is registered:

```
services.ConfigureAndValidate<MyOptions>(...)
```

The option's validation could be right on the class itself:

```
class MyOptions : IValidatableOption {
  public string A { get; set }

  public bool Validate() {
    if (string.IsNullOrEmpty(A)) throw new Exception();
  }
}
```

The `ConfigureAndValidate` extension method could do something like this under the hood:

```
services.Configure<MyOptions>(...);
services.Configure<MyOptions>(options => {
  options.Validate();
});
```

This way, `MyOptions` will be validated each time it is accessed. I'll go ahead and do these changes tonight so that you can properly validate your configurations.

LordMike:
@loic-sharma updated.

LordMike:
I also changed a few `ConfigureAndValidate<T>()` to `ConfigureAndValidateSection<T>()`

loic-sharma:
Could you rebase this off of the latest master? It's a little tricky to review right now with the line ending changes. Thanks!

LordMike:
I wasn't able to rebase, but I did copy all the changes to a new commit.

src/BaGet.AWS/BaGet.AWS.csproj
```diff
@@ -0,0 +1,18 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
loic-sharma: This project needs to be added to the solution file:

```
diff --git BaGet.sln BaGet.sln
index b487993..a9f015e 100644
--- BaGet.sln
+++ BaGet.sln
@@ -19,9 +19,11 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{26A0B557-53F
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "tests", "tests", "{C237857D-AD8E-4C52-974F-6A8155BB0C18}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "BaGet.Protocol", "src\BaGet.Protocol\BaGet.Protocol.csproj", "{A2D23427-9278-4D52-B31F-759212252832}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "BaGet.Protocol", "src\BaGet.Protocol\BaGet.Protocol.csproj", "{A2D23427-9278-4D52-B31F-759212252832}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "BaGet.Protocol.Tests", "tests\BaGet.Protocol.Tests\BaGet.Protocol.Tests.csproj", "{AC764A9A-9EAF-422B-9223-D3290C3CFD79}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "BaGet.Protocol.Tests", "tests\BaGet.Protocol.Tests\BaGet.Protocol.Tests.csproj", "{AC764A9A-9EAF-422B-9223-D3290C3CFD79}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "BaGet.AWS", "src\BaGet.AWS\BaGet.AWS.csproj", "{BA42BACD-1479-464E-BA04-07D156E80BAD}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
@@ -61,6 +63,10 @@ Global
 		{AC764A9A-9EAF-422B-9223-D3290C3CFD79}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{AC764A9A-9EAF-422B-9223-D3290C3CFD79}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{AC764A9A-9EAF-422B-9223-D3290C3CFD79}.Release|Any CPU.Build.0 = Release|Any CPU
+		{BA42BACD-1479-464E-BA04-07D156E80BAD}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{BA42BACD-1479-464E-BA04-07D156E80BAD}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{BA42BACD-1479-464E-BA04-07D156E80BAD}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{BA42BACD-1479-464E-BA04-07D156E80BAD}.Release|Any CPU.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
@@ -74,6 +80,7 @@ Global
 		{892A7A82-4283-4315-B7E5-6D5B70543000} = {C237857D-AD8E-4C52-974F-6A8155BB0C18}
 		{A2D23427-9278-4D52-B31F-759212252832} = {26A0B557-53FB-4B9A-94C4-BCCF1BDCB0CC}
 		{AC764A9A-9EAF-422B-9223-D3290C3CFD79} = {C237857D-AD8E-4C52-974F-6A8155BB0C18}
+		{BA42BACD-1479-464E-BA04-07D156E80BAD} = {26A0B557-53FB-4B9A-94C4-BCCF1BDCB0CC}
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
 		SolutionGuid = {1423C027-2C90-417F-8629-2A4CF107C055}
```

src/BaGet.AWS/Configuration/S3StorageOptions.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System.ComponentModel.DataAnnotations;
+using BaGet.Core.Validation;
+
+namespace BaGet.AWS.Configuration
+{
+    public class S3StorageOptions
+    {
+        [RequiredIf(nameof(KeySecret), null, IsInverted = true)]
+        public string KeyId { get; set; }
```
loic-sharma: It seems like `AccessKey` and `SecretKey` are the names used by the AWS SDK for `KeyId` and `KeySecret`. Could we use their terminology instead?

src/BaGet.AWS/BaGet.AWS.csproj
loic-sharma: You may need changes from the latest master for the CI to pass.

loic-sharma:
This is looking good. I left some final feedback, once that's addressed I'll merge this in :)

src/BaGet.AWS/BaGet.AWS.csproj
LordMike: Woops

loic-sharma:
Thank you @LordMike for the great work! 🎉

PR #145, opened 2018-12-03
Add new storage abstractions

loic-sharma:
See this conversation: https://github.com/loic-sharma/BaGet/pull/144#issuecomment-443544711

Adds a new interface, `IStorageService` for low level storage APIs. This interface is implemented by `FileStorageService` and `BlobStorageService` for Azure. All package content operations are now managed by `PackageStorageService` regardless of the low-level storage implementation.

/cc @LordMike


LordMike:
Other notes on service:

* Retries .. just re-run the command in the `PackageStorageService` on exceptions? -- or maybe just not handle it right now :P
* S3 has a possibility to delete multiple blobs in one go ... but it's tricky to support cases like that in generic interfaces.

To reap those benefits, I would either:

* Create a new interface, `IMultiDeletionCapableStorageService` (unrelated to `IStorageService`). `PackageStorageService` can then test if the service is that, and use its delete instead of the original `IStorageService`.
* Change the interface to always be multi-delete. Other implementations will then have to loop over the keys/paths provided to delete them indivdiually... This could be error prone, as every implementation now has to handle their individual errors, and they won't benefit from any retry-logic in `PackageStorageService`.

loic-sharma:
> Should the CancellationToken be passed all the way to the `IStorageService` ?

As in, is it necessary for the `IStorageService` methods to accept cancellation tokens? I agree that cancellation tokens aren't very useful for the `FileStorageService`, but the `BlobStorageService` may be doing considerable work over the network. I think there's value in being able to cancel those (and from experience, Azure's blob storage APIs can be shaky). Let me know if I misunderstood your question!

> Retries .. just re-run the command in the PackageStorageService on exceptions? -- or maybe just not handle it right now :P

Short answer: this is a surprisingly hard problem, and like you said, replaying the deletion command will fix up your state. I would rather tackle take this later.

Long answer: It is difficult to guarantee that the storage and the database are in sync. For example, what should happen if the database commit fails after a package was uploaded? Or, what happens if a package deletion fails half way through? Ideally, all these operations should appear atomic: the package should exist, or it shouldn't. Right now, BaGet can get in inconsistent states where the package can be downloaded because it exists in storage, but getting its metadata will fail because the package is not in the database. So what are some possible solutions? That I can think of, we can:

1. Use phased commits across storage and database. This greatly complicates things, I'd rather avoid this.
2. Make package operations atomic across storage and database content. This is possible if you also use your store package contents in the database, or...
3. Make the storage content-addressable. Upload/deletion operations now need a database lookup to find the correct content to act upon

So... this needs more thought :)

loic-sharma:
@LordMike feel free to leave any additional feedback here, I'll address them in a new PR

LordMike:
> > Should the CancellationToken be passed all the way to the `IStorageService` ?
> 
> As in, is it necessary for the `IStorageService` methods to accept cancellation tokens? I agree that cancellation tokens aren't very useful for the `FileStorageService`, but the `BlobStorageService` may be doing considerable work over the network. I think there's value in being able to cancel those (and from experience, Azure's blob storage APIs can be shaky). Let me know if I misunderstood your question!

I must've mistyped. I remembered the new `IStorageService` as _not_ getting the cancellation token. It does. So this point is moot. (I also should've written "shouldn't the ..") :P

PR #146, opened 2018-12-04
misc fixes

ku3mich:


src/BaGet.Azure/BaGet.Azure.csproj
```diff
@@ -1,19 +1,19 @@
...
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
+
+    <Description>The libraries to host BaGet on Azure.</Description>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.Search" Version="5.0.2" />
```
loic-sharma: Could you explain the need for these dependency changes?

src/BaGet/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -90,11 +90,12 @@ public static IServiceCollection AddBaGetContext(this IServiceCollection service
 
             services.AddDbContext<SqlServerContext>((provider, options) =>
             {
+
                 var databaseOptions = provider.GetRequiredService<IOptions<BaGetOptions>>()
                     .Value
                     .Database;
 
-                options.UseSqlServer(databaseOptions.ConnectionString);
+                options.UseSqlServer(databaseOptions.ConnectionString, t => t.UseRowNumberForPaging());
```
loic-sharma: Could you say more about the motivations for this change? It seems that this is used primarily for SQL Server 2008 support, but it also seems that EF Core has had bugs around this mode. I think I would rather avoid this option unless there's a very strong reason to add it in.

loic-sharma: Also, please remove the extra new line on line 93.

src/BaGet/Entities/AbstractContext.cs
```diff
@@ -33,6 +33,7 @@ public AbstractContext(DbContextOptions<TContext> options)
 
         protected override void OnModelCreating(ModelBuilder builder)
         {
+
```
loic-sharma: Please remove the extra new line here

src/BaGet/Migrations/Sqlite/20180907104341_Initial.Designer.cs
```diff
@@ -9,7 +9,7 @@
 namespace BaGet.Migrations.Sqlite
 {
     [DbContext(typeof(SqliteContext))]
-    [Migration("20180804082808_Initial")]
+    [Migration("20180907104341_Initial")]
```
LordMike: Renaming the migration seems like a bad idea, no?

@loic-sharma

loic-sharma: Good catch! This should be rolled back

src/BaGet.Azure/BaGet.Azure.csproj
ku3mich: the most of dependency changes was made because of project was unable to build at that time + upgrade packages to the latest. I did not control these changes, but thanks, will check and fix

src/BaGet/Extensions/ServiceCollectionExtensions.cs
ku3mich: yes, this change is specially for mssql 2008(R2 SP3 I use), EF generate queries with 'offset ... fetch' which is not supported by 2008, this some sort of quick fix this bug. and sure, I will remove

src/BaGet/Migrations/Sqlite/20180907104341_Initial.Designer.cs
ku3mich: sorry, u r right, don't understand how it is happen (

src/BaGet.Azure/BaGet.Azure.csproj
ku3mich: >> Microsoft.Azure.Storage.Blob v9.4.1 to WindowsAzure.Storage v9.3.1
seems it was installed a pre-release, I try to use only stable versions, so my opinion is to do not use pre-releases without a strong need, for now the latest stable is 9.3.3 and starting from 9.4 package contains a note:
>> NOTE: As of version 9.4.0, this library has been split into multiple parts and replaced:  See ...

I assume this was a reason of a failed build.

My suggestion is to stay with 9.3 for now

src/BaGet/Migrations/Sqlite/20180907104341_Initial.Designer.cs
ku3mich: I changed it back, but do not think it is right... at first: it is auto-generated file, 2nd: I see you(or scripts) made changes(add fields) to it rather than generate a new migration - it is not a right way to work with migrations

src/BaGet.Azure/BaGet.Azure.csproj
loic-sharma: What were the build issues you saw?

ku3mich:
done with your comments, put back blob 9.4, sorry I was wrong, updated nugets

src/BaGet.Azure/BaGet.Azure.csproj
ku3mich: I do not remember it was in September, then I have other work and and back to Nugets :) Version I have with my changes I use in our dev environment, seems works perfect - Thanks a lot for this project. I just want to bring something from myself to it )

loic-sharma:
Woah, this got huge. Can we split this up into manageable chunks? Also, please don't pull in changes from other branches into this one.

src/BaGet.Azure/BaGet.Azure.csproj
```diff
@@ -7,13 +7,17 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.Search" Version="5.0.1" />
+    <PackageReference Include="Microsoft.Azure.Search" Version="5.0.3" />
     <PackageReference Include="Microsoft.Azure.Storage.Blob" Version="9.4.1" />
-    <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="2.1.1" />
+    <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="2.2.0" />
```
loic-sharma: Please undo 2.2 changes. This will be merged by https://github.com/loic-sharma/BaGet/pull/149 when Azure DevOps supports .NET SDK 2.2.

src/BaGet.Core/BaGet.Core.csproj
```diff
@@ -1,20 +1,21 @@
...
+    <PackageReference Include="NuGet.Packaging" Version="4.9.1" />
     <PackageReference Include="System.Reflection.Metadata" Version="1.6.0" />
   </ItemGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\BaGet.Protocol\BaGet.Protocol.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Update="Nerdbank.GitVersioning" Version="2.2.33" />
```
loic-sharma: Why was this added in? The [`Directoy.Build.props`](https://github.com/loic-sharma/BaGet/blob/master/src/Directory.Build.props) brings this in for all relevant projects. This will break the Docker image build.

loic-sharma:
Hey @ku3mich, this change is too big to review, please break this up into issues. Let's discuss each issue, and once we agree on a solution, let's tackle each one in a small pull request. If you'd like, we can instant message on BaGet's new [gitter channel](https://gitter.im/BaGetServer/community). Thanks!

PR #147, opened 2018-12-05
Refactor configuration validation

loic-sharma:
Configuration are now registered using `services.ConfigureAndValidate<MyOptions>(...)`. This verifies that the configs' values match the option's annotations.

/cc @LordMike


src/BaGet/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -118,42 +128,46 @@ public static IServiceCollection ConfigureHttpServices(this IServiceCollection s
...
+        public static IServiceCollection ConfigureStorageProviders(this IServiceCollection services)
         {
+            services.AddFileStorageService();
+            services.AddBlobStorageService();
+
             services.AddTransient<IPackageStorageService, PackageStorageService>();
 
-            var storageOptions = configuration.Get<BaGetOptions>().Storage;
-            switch (storageOptions.Type)
+            services.AddTransient<IStorageService>(provider =>
```
loic-sharma: @LordMike this goes back to registering all storage implementations to the dependency injection container, and then using `IStorageService` to decide which implementation to use. This ensures that options aren't used until after they've been validated.

PR #148, opened 2018-12-05
Make it possible to configure the API key without hashing.

tmds:
Passing passwords unhashed is a common technique in containers.
Such passwords are stored and shared in the container platform.


loic-sharma:
What do you think of removing the `ApiKeyHash` config altogether? ASP.NET has a pretty good story around secrets in configurations, so hashing should be unnecessary.

tmds:
If no one has requested it, I think it we can remove it.
You can add it back if someone has a strong case for it.
Do you want me to remove it? That will be a breaking change.

LordMike:
Does nugets protocol for API Keys seriously just hash the API key?

It provides nothing over just sending the API key as is - it's the same shared secret... :|

loic-sharma:
@LordMike NuGet sends the API key in plaintext over HTTPS, but stores hashes of the API key in the database. You can think of an API key like a username and password credential.

LordMike:
Oh good. Hashing it brings nothing to the transport. :P

loic-sharma:
Thanks @tmds!

PR #158, opened 2018-12-05
Add Type decompilation services

LordMike:


src/BaGet.Decompiler.Nupkg/NugetDecompilerService.cs
```diff
@@ -0,0 +1,57 @@
...
+            // TODO: Read pdb's
+            // TODO: Read xml docs
+
+            var res = new AnalysisNugetPackage();
+
+            foreach (var item in package.GetReferenceItems())
+            {
+                // TODO: Can a reference be multiple files?
+                var file = item.Items.Single();
+                var folder = file.Substring(0, file.Length - Path.GetFileName(file).Length - 1);
```
loic-sharma: I think you can use `Path.GetDirectoryName` here

src/BaGet.Decompiler.Nupkg/Objects/AnalysisNugetAssemblyConverter.cs
```diff
@@ -0,0 +1,24 @@
...
+            Mapper = new MapperConfiguration(config =>
+            {
+                config.CreateMissingTypeMaps = false;
+                config.CreateMap<AnalysisAssembly, AnalysisNugetAssembly>();
+            }).CreateMapper();
+        }
+
+        public static AnalysisNugetAssembly Convert(AnalysisAssembly source)
+        {
+            return Mapper.Map<AnalysisNugetAssembly>(source);
```
loic-sharma: What's the value in using `AutoMapper` here? I'm wondering if we can cut the dependency

src/BaGet.Decompiler/AssemblyDecompilerService.cs
```diff
@@ -0,0 +1,184 @@
...
+
+        public AnalysisAssembly AnalyzeAssembly(Stream assembly, Stream pdb = null, Stream documentationXml = null)
+        {
+            string tempDir = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
+            Directory.CreateDirectory(tempDir);
+
+            AnalysisAssembly res;
+            try
+            {
+                string assemblyFile = Path.Combine(tempDir, "Binary.dll");
```
loic-sharma: You should be able to use [`StreamExtensions.AsTemporaryFileStreamAsync`](https://github.com/loic-sharma/BaGet/blob/master/src/BaGet.Core/Extensions/StreamExtensions.cs#L22) here.

src/BaGet.Decompiler.Nupkg/NugetDecompilerService.cs
LordMike: Nope. It turns the slashes.

src/BaGet.Decompiler.Nupkg/Objects/AnalysisNugetAssemblyConverter.cs
LordMike: I wanted to keep a seperation between the assembly decompilation, and the Nupkg decompilation. But then I also wanted to add extra info to the decompiled assemblies, such as the framework they targeted - hence this inheriting member.

I suppose we could join the Nupkg feature directly into the decompilation, so that it's a core part of it. It would, however, mean that you could decompile an assembly, by itself, but not know which framework it targeted.

src/BaGet.Decompiler/AssemblyDecompilerService.cs
LordMike: Maybe, although:

* I need the file name (can probably get that from the FileStream)
* I'm convinced the PDB file needs the same name, but different extension, as the dll file. Hasn't been doublechecked though.

LordMike: I've updated to ILSpy's 4.x packages, and it seems like they've written everything from scratch. It's entirely possible that we can do everything in memory now.

LordMike:
So. In order to obtain signatures for members and types (like `public class SomeClass`), I upgraded the CSharp Decompiler project to the new prerelease, version 4.x. 

Doing so cut out Cecil for decompilation, and boom, major speedup. For practically the same code, NEST took 5+ minutes on the old versions, while it now takes a few seconds. So. That's cool :)

src/BaGet.Decompiler.Nupkg/Objects/AnalysisNugetAssemblyConverter.cs
LordMike: I've merged them.

src/BaGet.Decompiler/AssemblyDecompilerService.cs
LordMike: On-disk files are no longer required. Yay.

LordMike:
So, @loic-sharma, I've got the decompilation in place. There's a lot to do with regards to pulling out documentation and whatnot, but the core is in place.

How would a type-searching-source-code-something service look like?

Following `ISearchService`, I imagine:

* `Task IndexAsync(PackageId id, AnalysisNugetAssembly code)` -- indexes details for a specific Package
* `Task<AnalysisNugetAssembly> GetAsync(PackageId id)` -- fetches a single packages code by id
* `Task<AnalysisType> GetAsync(PackageId id, NugetFramework targetFramework, string type)` -- fetches a single packages code, for a single type, by id+framework+type
* `Task<IReadOnlyList<CodeSearchResult>> FindAsync(string query, List<NugetFramework> allowedFrameworks = null)` -- finds "something", be it entire assemblies or types/members. 

And `CodeSearchResult` would be:

* `PackageId`
* `TargetFramework`
* `ResultType` -- `Type`, `Member`

Perhaps do this as inherited types?

* `TypeSearchResult : CodeSearchResult`
  * `Type` -- e.g. `MyBaseClass`
  * `Display` -- Some C# string, like `public abstract class MyBaseClass<T> : IDisposable where T : IOtherInterface`
* `MemberSearchResult : CodeSearchResult`
  * `Type` -- e.g. `MyBaseClass`
  * `MemberType` -- `Constructor`, `Method`, `Property`, `Field`, `Event`
  * `Display` -- Some C# string, like `public void DoWork();`

LordMike:
@loic-sharma I was wondering. I'm going through the binaries in the nupkg using ReferencedItems(). What's the correct way? ... do we risk getting native binaries this way?

loic-sharma:
This is looking great! Some comments:

1. The database migration is a little scary - once we add that in, we can't go back without breaking users that have migrated. Could we get merge the decompilation piece separately from the storage pieces?
2. Let's keep the decompilation separate from `BaGet.Core`. Can we keep this in a separate project named `BaGet.Decompilation` or something? I'm not sure how Entity Framework's models will fit in, but we can figure this out later!

I'd like to prototype running Roslyn analyzers on decompiled sources to find bugs in NuGet packages (incorrect `async` usage, etc...). This would be a great addition to [NuGet Package Explorer](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/)'s package analysis feature, so we should aim to make this as reusable as possible. Thanks for the great effort! :)

src/BaGet.Core/Decompiler/AssemblyDecompilerService.cs
```diff
@@ -0,0 +1,139 @@
...
+            //    pdbPe = new PEFile("Binary.pdb", pdb, PEStreamOptions.PrefetchEntireImage);
+
+            var decompiler = new CSharpDecompiler(assemblyPe, new NullAssmblyResolver(), new DecompilerSettings(LanguageVersion.Latest));
+            var res = DecompileAssembly(decompiler.TypeSystem.MainModule);
+
+            // Read documentation
+            // TODO Read documentation
+
+            // Fetch sources
+            localSourceCodeProviders.Add(new SourceCodeDecompiler(decompiler));
```
loic-sharma: Why is there a distinction between global and local source code providers? Could `SourceCodeDecompiler` be part of `_globalSourceCodeProviders `?

src/BaGet.Core/Decompiler/Decompilation/NullAssmblyResolver.cs
```diff
@@ -0,0 +1,17 @@
+using ICSharpCode.Decompiler.Metadata;
+
+namespace BaGet.Core.Decompiler.Decompilation
+{
+    internal class NullAssmblyResolver : IAssemblyResolver
```
loic-sharma: This is missing an "e": `NullAssemblyResolver`

src/BaGet.Core/Decompiler/SourceCode/SourceCodeDecompiler.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace BaGet.Core.Decompiler.SourceCode
+{
+    internal class SourceCodeDecompiler : ISourceCodeProvider
+    {
+        private readonly CSharpDecompiler _decompiler;
+
+        public SourceCodeDecompiler(CSharpDecompiler decompiler)
+        {
+            _decompiler = decompiler;
```
loic-sharma: Please add null checks to the constructor: `_decompiler = decompiler ?? throw new ArgumentNullException(nameof(decompiler));`

src/BaGet.Core/Decompiler/SourceCode/ISourceCodeProvider.cs
```diff
@@ -0,0 +1,10 @@
+using BaGet.Core.Entities;
+using ICSharpCode.Decompiler.TypeSystem;
+
+namespace BaGet.Core.Decompiler.SourceCode
+{
+    internal interface ISourceCodeProvider
+    {
+        bool TryFillSources(IModule module, SourceCodeAssembly assembly);
```
loic-sharma: Instead of accepting a `SourceCodeAssembly`, could this accept a `SourceCodeType`?  Also, could we make it so that `ISourceCodeProvider.TryFillSources` returns `true` only if `SourceCodeType.CSharp` was filled in?

With these changes, `AssemblyDecompilerService` would play a little nicer if there are multiple source code providers

src/BaGet.Core/Decompiler/AssemblyDecompilerService.cs
LordMike: I initially planned this as a series of "tryDecompileSource" methods, where Globals can run for all decompilations while Locals are specific for this one run of decompilation. We try not to create too many objects where I come from :).

The whole transient-preferred mindset is new to me. In that light, every decompiler is "local" (create pr. use), so this global thing can go.

src/BaGet.Core/Decompiler/SourceCode/ISourceCodeProvider.cs
LordMike: True. I'm a bit worried about how to find sources from embedded / linked sources, since there may not be a 1:1 relationship between files and types... But that may be resolved later.

LordMike:
As I mentioned, I tried the Baget.Decompilation route, but quickly realized this thing with the entities. So to not duplicate all those (one for DB stuff and one for Decompilation), I chose this route.

However. It may be that there _should_ be a DB entity and a not-db-entity. So that the DB entity can include relationships needed for the DB, while the Decompilation entity can include only decompilation-related stuff. Then it's a non-issue to move out decompilation (again :)); at the cost of mapping between these two entity types.

How'd that sound?

LordMike:
Also. There is a slight space issue. Having uploaded one NEST assembly (3.400 public types), the Sqlite DB grew to 24 MB... Soo.. that's a lot of space within a short timespan.

loic-sharma:
The mapping between decompilation objects and database entities sounds good to me. I’m wondering if maybe we should only store type and method information in the database? Would that reduce storage costs if we cut out the C#? We’d have to regenerate the C# code everytime it is requested.

I’ll be out for the holidays with no internet, so I won’t be able to review until the end of next week.

LordMike:
Decompiling on the fly is a no go... The packages would have to be downloaded for each view.. A few options:

* Compressing the C# code in the DB (it's not for searching anyways)
* Decompiling the C# later on (decompile _all_ types in one go, when _any_ type is wanted, store the results in DB)
* Store the C# elsewhere (compressed json f.ex., perhaps on the same storage as packages, alternatively on a separate storage)

LordMike:
Decompiling on the fly is a no go... The packages would have to be downloaded for each view.. A few options:

* Compressing the C# code in the DB (it's not for searching anyways)
* Decompiling the C# later on (decompile _all_ types in one go, when _any_ type is wanted, store the results in DB)
* Store the C# elsewhere (compressed json f.ex., perhaps on the same storage as packages, alternatively on a separate storage)
  * Cache this on presentation, so that it can be shown multiple times within f.ex. an hour, ..

loic-sharma:
Note for myself, look into https://github.com/KirillOsenkov/SourceBrowser

LordMike:
That could be cool - dump the members for searching purposes, then use this for display.

loic-sharma:
That's what I was thinking too. @LordMike could we IM on [gitter](https://gitter.im/BaGetServer/community#) about this feature? If so, when would be a good time for you? I wanna make sure we're on the same page 😄

PR #162, opened 2018-12-06
FileStorageService: verify path resolves under the store directory

tmds:


src/BaGet.Core/Services/FileStorageService.cs
```diff
@@ -96,13 +99,19 @@ public Task DeleteAsync(string path, CancellationToken cancellationToken = defau
...
             {
                 throw new ArgumentException("Path is required", nameof(path));
             }
 
-            return Path.Combine(_storePath, path);
+            // Verify path is under the _storePath.
+            string fullPath = Path.GetFullPath(Path.Combine(_storePath, path));
+            if (!fullPath.StartsWith(_storePath))
+            {
+                throw new ArgumentException("Path resolves outside store path");
```
loic-sharma: Could you add a `nameof(path)` to be consistent with line 104?

loic-sharma: Also please add a unit test in [`FileStorageServiceTests.cs`](https://github.com/loic-sharma/BaGet/blob/master/tests/BaGet.Core.Tests/Services/FileStorageServiceTests.cs)

loic-sharma:
Argh. It looks like the CI is broken for external contributors. I'll fix that tonight 😓

tests/BaGet.Core.Tests/Services/FileStorageServiceTests.cs
```diff
@@ -191,6 +234,25 @@ protected async Task<string> ToStringAsync(Stream input)
...
+                get
+                {
+                    string fullPath = Path.GetFullPath(_storePath);
+                    yield return "../file";
+                    yield return ".";
+                    yield return $"../{Path.GetFileName(_storePath)}";
+                    yield return $"../{Path.GetFileName(_storePath)}suffix";
+                    yield return $"../{Path.GetFileName(_storePath)}suffix/file";
+                    yield return fullPath;
+                    yield return fullPath + Path.DirectorySeparatorChar;
```
loic-sharma: This test case fails on Windows. See my comment on line 109 of `FileStorageService`.

src/BaGet.Core/Services/FileStorageService.cs
```diff
@@ -96,13 +99,19 @@ public Task DeleteAsync(string path, CancellationToken cancellationToken = defau
...
-            // TODO: This should check that the result is in _storePath for security.
             if (string.IsNullOrEmpty(path))
             {
                 throw new ArgumentException("Path is required", nameof(path));
             }
 
-            return Path.Combine(_storePath, path);
+            // Verify path is under the _storePath.
+            string fullPath = Path.GetFullPath(Path.Combine(_storePath, path));
+            if (!fullPath.StartsWith(_storePath))
```
loic-sharma: If the `fullPath` and `_storePath` are identical, this check will pass. This causes the test `fullPath + Path.DirectorySeparatorChar;` to fail. I would add something like:

```
if (fullPath.Equals(_storePath, StringComparison.InvariantCultureIgnoreCase))
{
    throw new ArgumentException("Path must be a file within the store path", nameof(path));
}
```

tests/BaGet.Core.Tests/Services/FileStorageServiceTests.cs
loic-sharma: I'm loving how thorough these tests are 👍

src/BaGet.Core/Services/FileStorageService.cs
tmds: I've added a length check. I also changed the startswith compare to an Ordinal one.

loic-sharma:
Thank you! :)

PR #171, opened 2018-12-10
Added env var to specify other appsettings

Mizipzor:
### What does this PR do?

By launching the app with environment variable `APPSETTINGS_ROOT` set that folder will be used to look for appsettings.json files.

### Motivation

The appsettings.json file contains sensitive information. We are hosting on Kubernetes and are mounting sensitive settings and certificates in a secret volume. This mount, and thus it files, cant go in the root, they need a subfolder. As such, I need a way to override where it looks for the settings file. There might be a more elegant way to do this.

### DIFF
--- a/src/BaGet/Program.cs
+++ b/src/BaGet/Program.cs
@@ -56,14 +56,9 @@ namespace BaGet
                 })
                 .ConfigureAppConfiguration((builderContext, config) =>
                 {
-                    var root = Environment.GetEnvironmentVariable("APPSETTINGS_ROOT");
+                    var root = Environment.GetEnvironmentVariable("BAGET_CONFIG_ROOT");
                     if (!string.IsNullOrEmpty(root))
-                    {
-                        var env = builderContext.HostingEnvironment;
-                        config.Sources.Clear();
-                        config.AddJsonFile($"{root}/appsettings.json", optional: false, reloadOnChange: true)
-                            .AddJsonFile($"{root}/appsettings.{env.EnvironmentName}.json", optional: true, reloadOnChange: true);
-                    }
+                        config.SetBasePath(root);
                 });
 
         public static IHostBuilder CreateHostBuilder(string[] args)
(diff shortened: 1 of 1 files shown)


src/BaGet/Program.cs
```diff
@@ -51,6 +53,17 @@ await provider
...
                     // be enforced by a reverse proxy server, like IIS.
                     options.Limits.MaxRequestBodySize = null;
+                })
+                .ConfigureAppConfiguration((builderContext, config) =>
+                {
+                    var root = Environment.GetEnvironmentVariable("APPSETTINGS_ROOT");
+                    if (!string.IsNullOrEmpty(root))
+                    {
+                        var env = builderContext.HostingEnvironment;
+                        config.Sources.Clear();
```
loic-sharma: What's the `Clear();` for?

src/BaGet/Program.cs
```diff
@@ -51,6 +53,17 @@ await provider
...
+                        config.AddJsonFile($"{root}/appsettings.json", optional: false, reloadOnChange: true)
```
loic-sharma: Did you try instead setting `config.SetBasePath(root);`? You may not need to re-add the json file names.

src/BaGet/Program.cs
Mizipzor: If you use the docker image unaltered it will already include appsettings at a place listed by default in `config.Sources`. So you either build the image yourself and make sure to remove those files before building, or you have this bit that clears the current settings before applying the new ones.

loic-sharma: I think a better alternative is to use `SetBasePath`. You don’t need to delete the old files and you can use your secret volume’s settings files.

src/BaGet/Program.cs
loic-sharma: If this approach works, can we rename `APPSETTINGS_ROOT` to `BAGET_CONFIG_ROOT`

src/BaGet/Program.cs
Mizipzor: I agree. Will have another look and get back to you.

loic-sharma:
Thanks!

PR #172, opened 2018-12-10
Added support for Google buckets storage.

Mizipzor:
### What does this PR do?

Adds support for Google buckets storage.

### Motivation

Needed for persistent storage.

### DIFF
--- a/src/BaGet.Core/BaGet.Core.csproj
+++ b/src/BaGet.Core/BaGet.Core.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
@@ -15,6 +15,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\BaGet.Protocol\BaGet.Protocol.csproj" />
+    <PackageReference Include="Google.Cloud.Storage.V1" Version="1.0.0" />
   </ItemGroup>
 
 </Project>
--- a/src/BaGet.Core/BaGet.Core.csproj
+++ b/src/BaGet.Core/BaGet.Core.csproj
@@ -15,7 +15,6 @@
 
   <ItemGroup>
     <ProjectReference Include="..\BaGet.Protocol\BaGet.Protocol.csproj" />
-    <PackageReference Include="Google.Cloud.Storage.V1" Version="1.0.0" />
   </ItemGroup>
 
 </Project>
--- a/src/BaGet.Core/Services/GoogleBucketStorageService.cs
+++ b/src/BaGet.Core/Services/GoogleBucketStorageService.cs
@@ -1,0 +1,54 @@
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using BaGet.Core.Configuration;
+using Google.Cloud.Storage.V1;
+using Microsoft.Extensions.Options;
+
+namespace BaGet.Core.Services
+{
+    public class GoogleBucketStorageService : IStorageService
+    {
+        private readonly string _bucketName;
+
+        public GoogleBucketStorageService(IOptionsSnapshot<GoogleBucketStorageOptions> options)
+        {
+            if (options == null)
+                throw new ArgumentNullException(nameof(options));
+
+            _bucketName = options.Value.BucketName;
+        }
+
+        public async Task<Stream> GetAsync(string path, CancellationToken cancellationToken = default)
+        {
+            var storage = await StorageClient.CreateAsync();
+            var stream = new MemoryStream();
+            await storage.DownloadObjectAsync(_bucketName, path, stream, cancellationToken: cancellationToken);
+            stream.Seek(0, SeekOrigin.Begin);
+            return stream;
+        }
+
+        public Task<Uri> GetDownloadUriAsync(string path, CancellationToken cancellationToken = default)
+        {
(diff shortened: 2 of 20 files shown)


src/BaGet.Core/BaGet.Core.csproj
```diff
@@ -15,6 +15,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\BaGet.Protocol\BaGet.Protocol.csproj" />
+    <PackageReference Include="Google.Cloud.Storage.V1" Version="1.0.0" />
```
loic-sharma: Please create a new project `BaGet.Gcp.csproj` and move `GoogleBucketStorageOptions` and `GoogleBucketStorageService` there

loic-sharma: See https://github.com/loic-sharma/BaGet/pull/144 as an example

src/BaGet.Core/Services/GoogleBucketStorageService.cs
```diff
@@ -0,0 +1,54 @@
...
+            return Task.FromResult(new Uri(path));
+        }
+
+        public async Task<PutResult> PutAsync(string path, Stream content, string contentType, CancellationToken cancellationToken = default)
+        {
+            var storage = await StorageClient.CreateAsync();
+            var obj = await storage.GetObjectAsync(_bucketName, path, cancellationToken: cancellationToken);
+            if (obj != null)
+                return PutResult.AlreadyExists;
+            await storage.UploadObjectAsync(_bucketName, path, null, content, cancellationToken: cancellationToken);
```
loic-sharma: Why `null`? Can you pass the `contentType` instead?

Mizipzor:
Named the new project `BaGet.GCP` rather than `BaGet.Gcp` to match capitalization of aws project, I'm not sure which you prefer.

src/BaGet.GCP/BaGet.GCP.csproj
```diff
@@ -0,0 +1,24 @@
...
+  <ItemGroup>
+    <PackageReference Include="Google.Cloud.Storage.V1" Version="2.2.1" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\BaGet.Core\BaGet.Core.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Reference Include="Microsoft.Extensions.DependencyInjection.Abstractions">
```
loic-sharma: Please use `PackageReference` here

src/BaGet.GCP/Services/GoogleBucketStorageService.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            if (options == null)
+                throw new ArgumentNullException(nameof(options));
+
+            _bucketName = options.Value.BucketName;
+        }
+
+        public async Task<Stream> GetAsync(string path, CancellationToken cancellationToken = default)
+        {
+            var storage = await StorageClient.CreateAsync();
```
loic-sharma: `StorageClient` is disposable, please wrap all `StorageClient.CreateAsync()` with `using`s.

src/BaGet.GCP/Services/GoogleBucketStorageService.cs
```diff
@@ -0,0 +1,65 @@
...
+            var stream = new MemoryStream();
+            await storage.DownloadObjectAsync(_bucketName, path, stream, cancellationToken: cancellationToken);
+            stream.Seek(0, SeekOrigin.Begin);
+            return stream;
+        }
+
+        public Task<Uri> GetDownloadUriAsync(string path, CancellationToken cancellationToken = default)
+        {
+            return Task.FromResult(new Uri(path));
```
loic-sharma: This returns the incorrect Uri. Two suggestions:

1. Use [`UrlSigner.Sign`](https://www.fuget.org/packages/Google.Cloud.Storage.V1/2.3.0-beta04/lib/netstandard1.3/Google.Cloud.Storage.V1.dll/Google.Cloud.Storage.V1/UrlSigner) to create a signed Uri to the object. This seems hard to do...
2. Assume that the URI is public and return `http://storage.googleapis.com/BUCKET_NAME/OBJECT_NAME`

I would go for the latter.

src/BaGet.GCP/Services/GoogleBucketStorageService.cs
```diff
@@ -0,0 +1,85 @@
...
+            return Task.FromResult(new Uri($"http://storage.googleapis.com/{_bucketName}/{CoercePath(path).TrimStart('/')}"));
+        }
+
+        public async Task<PutResult> PutAsync(string path, Stream content, string contentType, CancellationToken cancellationToken = default)
+        {
+            using (var storage = await StorageClient.CreateAsync())
+            {
+                try
+                {
+                    await storage.GetObjectAsync(_bucketName, CoercePath(path), cancellationToken: cancellationToken);
```
loic-sharma: This isn't quite right. The steps should be something like:

1. Attempt to `UploadObjectAsync`
2. If the upload fails because the file exists (likely a `HttpStatusCode.Conflict`, but please verify this as I've never used GCP), then:
    1. Download the object
    2. Return `PutResult.AlreadyExists` if the object is the same as `content`, or `PutResult.Conflict` otherwise.

For an example of this, please see Azure's [`BlobStorageService`](https://github.com/loic-sharma/BaGet/blob/master/src/BaGet.Azure/BlobStorageService.cs#L68)

bgrainger:
@loic-sharma This has been superseded by #233 and can be closed.

loic-sharma:
Yes it can, thanks for the reminder! :)

Issue #3, opened 2018-03-14
Implement package drop folder with automatic indexing


loic-sharma:
Allow BaGet to automatically index packages that are dropped in a designated folder.

See [NuGet.Server's `ServerPackageRepository`](https://github.com/NuGet/NuGet.Server/blob/2ee751ab71048d87f97f45807a420517a9aa9e27/src/NuGet.Server.Core/Infrastructure/ServerPackageRepository.cs).

LordMike:
Similar to #72 -- imagine a queue service being based on a directory. It could then poll the directory and kick off a task/delegate for each item in the directory.

badre429:
@loic-sharma  any update ?

loic-sharma:
@badre429 A community member started this work in https://github.com/loic-sharma/BaGet/pull/174, but I haven't had the chance to complete it. Out of curiosity, what's your scenario?

badre429:
@loic-sharma  my company has offline build policy there fore we are looking for replacement to nuget server working on linux servers

kirichenec:
Still in progress or not?))

PR #174, opened 2018-12-11
Added support for re-indexing all missing packages on startup.

Mizipzor:
### What does this PR do?

Adds support for, on startup, to go through all packages in the store and re-adds them to the database.

### Motivation

The intention is to have the option to treat the sqllite database living inside the container as a throwaway cache, since the container is rarely restarted and its one less connection to secure.

### Additional Notes

Could use some more work, but I wanted some feedback on this feature and the general approach. One thing of note is that the downloads count for packages would be reset if the docker container is restarted (something we dont care about).


loic-sharma:
I like the idea of being able to use the SQLite database as a throwaway cache. However, the current implementation feels a little hacky. I wonder if we can kill two birds with one stone by implementing a "drop folder" with automatic indexing (see https://github.com/loic-sharma/BaGet/issues/3), similar to [NuGet.Server's `ServerPackageRepository`](https://github.com/NuGet/NuGet.Server/blob/2ee751ab71048d87f97f45807a420517a9aa9e27/src/NuGet.Server.Core/Infrastructure/ServerPackageRepository.cs).

For your case, the "drop folder" would be the `packages` folder. Any unknown packages would automatically be indexed at startup.

src/BaGet.Core/Services/PackageIndexingService.cs
```diff
@@ -32,7 +32,12 @@ public class PackageIndexingService : IPackageIndexingService
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
         }
 
-        public async Task<PackageIndexingResult> IndexAsync(Stream packageStream, CancellationToken cancellationToken)
+        public async Task<PackageIndexingResult> IndexAsync(
+            Stream packageStream,
+            bool storage = true,
```
loic-sharma: Please remove the booleans you added. `IndexAsync` should be idempotent. If you call `IndexAsync` for a package you've already indexed, absolutely nothing should happen: it should no-op without throwing any exceptions. For example, `FileStorageService` allows you to upload the same package to storage twice (see [this code](https://github.com/loic-sharma/BaGet/blob/master/src/BaGet.Core/Services/FileStorageService.cs#L76)). This means that you should be able to blindly call `IndexAsync` on any package you find in the drop folder. Some notes:

1. Currently, idempotency is broken on AWS S3 and GCP. We should fix that...
2. Calling `IndexAsync` twice on the same package should "reindex" the package and reupload it to storage and reindex it in search. I wouldn't worry about this now as this should be out of scope for this pull request

src/BaGet.Core/Services/PackagesReconciler.cs
```diff
@@ -0,0 +1,41 @@
...
+            ILogger<PackageIndexingService> logger)
+        {
+            _storage = storage;
+            _index = index;
+            _logger = logger;
+        }
+
+        public async Task Reconcile(CancellationToken cancellationToken = default)
+        {
+            var allStored = await _storage.GetPackagePathsAsync(cancellationToken);
```
loic-sharma: Let's not use `IStorageService` here, I'm okay with only supporting the local filesystem.

src/BaGet.Core/Services/PackagesReconciler.cs
```diff
@@ -0,0 +1,41 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace BaGet.Core.Services
+{
+    public class PackagesReconciler
```
loic-sharma: Let's rename this to `DropFolderService`

src/BaGet/Startup.cs
```diff
@@ -54,6 +55,15 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
                 }
             }
 
+            if (options.StartupEnsurePackagesExistsInDatabase)
+            {
+                using (var scope = app.ApplicationServices.CreateScope())
+                {
+                    var packagesReconciler = scope.ServiceProvider.GetService<PackagesReconciler>();
+                    packagesReconciler.Reconcile().Wait(); // TODO better place for long running start routines
```
loic-sharma: This should be a hosted service. See [this documentation](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/hosted-services?view=aspnetcore-2.2#timed-background-tasks)

src/BaGet.Core/Configuration/BaGetOptions.cs
```diff
@@ -21,6 +21,12 @@ public class BaGetOptions
         /// </summary>
         public bool RunMigrationsAtStartup { get; set; } = true;
 
+        /// <summary>
+        /// If enabled, the packages folder will be enumerated and every package
+        /// is ensured to exist in the database.
+        /// </summary>
+        public bool StartupEnsurePackagesExistsInDatabase { get; set; }
```
loic-sharma: Let's create a new option `AutomaticallyIndexedFolder` with two properties: `Enabled` and `Path`.

Mizipzor:
Agree on all points. This turned out a bit rushed and was only meant as a workaround since I couldnt get a proper database to work. It should be replaced with a proper implementation for a drop folder, but thats not something I have time for now. Should I just close this or leave it open for someone else to build upon?

loic-sharma:
Let’s leave this open as it’s 80% the way there. I’ll finish this up in the future. Thanks for starting this work!

rdubya16:
Is there any progress on this @Mizipzor @loic-sharma we are running Baget in kubernetes backed by S3. Every time a pod gets restarted we loose the metadata in the web UI. This would be a nice feature to have.

Mizipzor:
I'm afraid I had to abandon work on this due to time constraints. It's been left here in its current state for someone else to build upon, if it would save them time. I might return to it, should the opportunity presents itself.

PR #175, opened 2018-12-12
Added MySQL server support

Mizipzor:
### What does this PR do?

Adds MySQL server support.

### Motivation

Need it. Sort of gave up on getting PostgresSql to work.

### Additional Notes

Sadly the current entity framework provider doesnt support the `LIMIT` keyword in queries. So `DatabaseSearchService` couldnt fetch packages. Added another service that doesnt do the linq take or skip, which is currently not a great loss since the paging doesnt appear to be used anyway (?). Cant find an issue for this in their github repos, will try to find the time to reproduce this minimally and create one.


loic-sharma:
> Cant find an issue for this in their github repos, will try to find the time to reproduce this minimally and create one.

I would say this is a blocker for merging this change in as otherwise the search experience will be poor. Is this limit issue easy to reproduce using BaGet?

McSoda:
Hi,

May be the solution for this db provider is to made the paging outside the search query:

            var packages = await _context.Packages
                .Where(p =>
                    search.Select(p2 => p2.Id)
                        .OrderBy(id => id)
                        .Distinct()
                        .Contains(p.Id))
                .Skip(skip)
                .Take(take)
                .GroupBy(p => p.Id)
                .ToListAsync();

Regards

loic-sharma:
Yes something like that could be done. I suppose it's better to have MySQL support even if the performance isn't optimal. I'll see if I can get this merged in

McSoda:
Hi,

Concerning Mizipzor pull request, would probably:
* remove the unnecessary class DatabaseSearchServiceNoPaging
* add a property to IContext interface:
        bool IsLimitSupported { get; }
* implement the property in AbstractContext class:
        public virtual bool IsLimitSupported { get => true; }
* implement the property in MySqlContext class:
        public override bool IsLimitSupported { get => false; }
* modify the DatabaseSearchService class as explained in my previous message.

At a second time, some could replace the property IsLimitSupported by a more general InsupportedFeatures collection to manage differences between dbContext implementations in EF.

Regards,

loic-sharma:
@Mizipzor Thanks for getting this setup. I added @McSoda's suggestions and merged this in as part of https://github.com/loic-sharma/BaGet/pull/206. I kept your commit so that your get credit in the commit history. Cheers 😄

Issue #43, opened 2018-06-05
Support v2 APIs


loic-sharma:
Support OData endpoints. Use: https://www.nuget.org/packages/Microsoft.AspNetCore.OData

loic-sharma:
See this blog: https://blogs.msdn.microsoft.com/odatateam/2018/07/03/asp-net-core-odata-now-available/

WernerMairl:
How important is v2 at this time ?

Observing OData.net over longer time (and the very long/slow way to dotnet core) i‘m not really convinced that the effort here is a good investment....

loic-sharma:
Right now, it's not very important - I'd consider more of a "nice to have". Reasons for v2:

* To reach feature parity against https://github.com/NuGet/NuGet.Server
* To support very old NuGet clients

NN---:
There is LiGet with v2 .
Probably you can merge code bases.

https://github.com/ai-traders/liget

tomzo:
I have ported V2 implementation from LiGet to BaGet in [my fork](https://github.com/ai-traders/baget), you can try it.

Jaykul:
If you _do_ add this, perhaps make it an option that must be turned on, so that people don't inadvertently let their users become dependent on it, and get stuck keeping it around "forever" (the way nuget.org is). Let's not forget that [there were reasons](https://devblogs.microsoft.com/nuget/switching-from-wcf-odata-to-web-api/) why NuGet.org had to toss their original [OData implementation](https://devblogs.microsoft.com/nuget/nuget-architecture/) 😉

loic-sharma:
> If you do add this, perhaps make it an option that must be turned on

I was strongly considering this. I think this may be a poor experience for Powershell users though, so I think for now I'll enable it by default.

>  get stuck keeping it around "forever" (the way nuget.org is). Let's not forget that there were reasons why NuGet.org had to toss their original OData implementation 😉

Oh boy, I know. I'm on the team that runs nuget.org, and the OData APIs are a real thorn in our side. They don't scale nearly as well as the new V3 APIs and it is much, much more expensive even with significantly lower traffic. It'd be great if everyone could move to V3 right away! 😅

BlythMeister:
I am migrating from Klondike nuget feed to baget.
Klondike was a V2 endpoint at "/API/odata"
[...]

PR #178, opened 2018-12-13
Added support for acting like a Chocolatey server.

Mizipzor:
### What does this PR do?

Support for [Chocolatey](https://chocolatey.org/).
This includes rudimentary support for api v2 and OData.
Lifted a lot of the core logic from https://github.com/ai-traders/BaGet.
Created Legacy namespace and tried to keep everything contained, would prefer if Choco added nuget api v3 support.

### Closes Issue(s)

#43 

At least partially, OData queries can be very expressive, I dont know to what extent old nuget clients use it.

### Motivation

Need it, and want to reuse the same server software between nuget and choco.

### Additional Notes

Would like some help in structuring this in a better way, support for api v2 got added but code became more complex. Not sure if its worth the cost. But the lack of support turned out to be a deal breaker for us.

### DIFF
--- a/src/BaGet/BaGet.csproj
+++ b/src/BaGet/BaGet.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk.Web">
+<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
     <TargetFramework>netcoreapp2.2</TargetFramework>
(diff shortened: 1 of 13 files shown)


Mizipzor:
Should have looked more before I started, this is the class I should have mimicked:

https://github.com/NuGet/NuGet.Server/blob/master/src/NuGet.Server.V2/Controllers/NuGetODataController.cs

While porting it to dotnet core is probably non-trivial, it should at least show proper mapping of the urls.

mycroes:
@loic-sharma Is there anything holding you back from merging this in it's current state? Just like @Mizipzor I want to use BaGet for Chocolatey as well (besides being our actual NuGet package source) and I think it would be easier to expose V2 api in BaGet than waiting for Chocolatey V3 support.

I'd also love to hear if you're not looking forward to adding V2 support at all. Then it would perhaps be better to create a separate (as in, not tied to BaGet in any way) V2 proxy just to enable Chocolatey access to BaGet.

WesleyKhoiVo:
Hi @loic-sharma, Is there anything prevent you merge this feature?

PR #179, opened 2018-12-18
Take advantage of docker cache

phaza:
By moving these two RUNs above COPY . ., docker will be able to cache the installation of nodejs and speed up builds.


loic-sharma:
Looks great! I’ll test this next week when I’m back from the holidays and merge this in. Thanks! :)

loic-sharma:
I apologize for taking so long for merging this in. Thanks!

PR #183, opened 2018-12-20
Bug fix: link for packages don't work after input something for search

vitaliy-orlov:
Hi!

How to reproduce a bug:
1. Input something to search component
2. Click on package

## Expected Behavior

View package detailed information

## Actual Behavior

Nothing happens. Only changed url in browser.

PR #186, opened 2018-12-30
Fix navigation after searching

loic-sharma:
* Fix navigation after the user has inputted search content
* Fixed some issues where `setState` was called after the component was unmounted
* Cleaned up routing

This was reported and originally fixed by @vitaliy-orlov in https://github.com/loic-sharma/BaGet/pull/183.


src/BaGet.UI/src/App.tsx
```diff
@@ -1,35 +1,40 @@
...
-          {this._renderContent()}
-        </div>
-      </Router>
+        {this._renderContent()}
+      </div>
     );
   }
 
+  private onRouteChange = (location: Location, action: Action) =>
+    this.setState({ input: "" });
```
loic-sharma: @vitaliy-orlov this is the fix

PR #202, opened 2019-02-02
PR: Refactorings for RegistrationIndex

WernerMairl:
- RegistrationIndex logic refactored from Controller into Service/IService
- Including test skeleton for calling RegistrationIndexController via xUnit (ControllerTests)


loic-sharma:
I like this idea, but I want to reflect more on how the URL generation should work. Ideally the URL generation service should be injected into dependencies, maybe that's possible using ASP.NET Core's new [`LinkGenerator`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.routing.linkgenerator?view=aspnetcore-2.2). I need to do a little more research on this front, but thanks for getting the ball rolling here!

Issue #227, opened 2019-03-07
Add support for HTTPS


mu88:
Right now, BaGet is always running under `http://localhost:5000`. It would be great to have a switch in `appsettings.json` to run under HTTPS.

WernerMairl:
BaGet is working fine with https - depending on the hosting area.
(Azure WebApp/Kudu, Docker Environment...) 

Appsettings.json is only the development default, you can/should define your requirements on runtime!

Yes, there are options to implement https inside Kestrel, but in my experience typically this the "part of the job" that should be implemented by the hosting environment and not inside BaGet.

If you provide more information what you try to do, we may be able to help you in some way...

Werner

mu88:
In most of the cases, I agree with you, Werner. However, for internal purposes, I don't need a full-fledged web server. I want a simple 'download and run' NuGet server without the need to setup all the other stuff. That's what I love about BaGet!

loic-sharma:
I personally use the same approach that @WernerMairl describes and this works well for production workloads. That being said I completely agree with @mu88, BaGet should work well as a standalone app. I'll go ahead and a config for this, thanks for the suggestion! :)

cabal95:
@WernerMairl Can you explain how to make SSL work, since you mention that it already works? My environment is a Linux server with BaGet running in a docker container. In order to get SSL I have to use a reverse proxy that handles the SSL itself. However, BaGet builds absolute URLS based solely on the request URL (which means it becomes `localhost:5555`) so the `/v3/index.json` response comes back with nothing but `localhost:5555` URLs, which obviously causes package submission to fail. There doesn't seem to be a way to override the hostname/port used.

hypervtechnics:
@cabal95 As far as I can see the Urls generated by BaGet take there information from the Forward headers. Did you set them correctly. Also there is the PathBase setting if you are running in subdirectory.

cabal95:
Yes, it is. I just verified with a tcpdump trace that Apache is sending along the `X-Forwarded-` headers. I'm just using Apache's standard reverse proxy config:

```
	<Location />
		ProxyPass http://localhost:5555/
		ProxyPassReverse http://localhost:5555/
	</Location>
```

However, it is partially working now. Yesterday I found and added the `ProxyPreserveHost On` option (which from the docs means Apache sends the original Hostname requested instead of the "new" hostname) and that seemed to fix it from sending back `localhost` instead of the proper hostname.

[...]

PR #231, opened 2019-03-10
Add support for HTTPS #227

mu88:
Allows to use Kestrel with HTTPS even without a Reverse Proxy (e. g. for development or internal purposes).

Addresses https://github.com/loic-sharma/BaGet/issues/227

See https://github.com/loic-sharma/BaGet/issues/227

 * Kestrel configuration was modified so that it can use HTTPS if configured.

### DIFF
--- a/src/BaGet.Core/Configuration/HttpsOptions.cs
+++ b/src/BaGet.Core/Configuration/HttpsOptions.cs
@@ -1,0 +1,11 @@
+namespace BaGet.Core.Configuration
+{
+    public class HttpsOptions
+    {
+        public int Port { get; set; }
+
+        public string CertificateFileName { get; set; }
+
+        public string CertificatePassword { get; set; }
+    }
+}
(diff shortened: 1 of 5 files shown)


src/BaGet/appsettings.json
```diff
@@ -1,5 +1,5 @@
 {
-  "ApiKey": "",
+  "ApiKey": "SGD1953",
```
loic-sharma: Please revert this. I would recommend generating a new API key if this is what you use on your instance of BaGet

mu88:
@loic-sharma I hope, the PR is now okay for you. If not, please do not hesitate to contact me :)

src/BaGet.Core/Configuration/HttpsOptions.cs
```diff
@@ -0,0 +1,11 @@
...
+{
+    public class HttpsOptions
+    {
+        public int Port { get; set; }
+
+        public string CertificateFileName { get; set; }
+
+        public string CertificatePassword { get; set; }
+    }
+}
```
mickvangelderen, line 11: I don't understand the need to add a custom options class. Doesn't kestrel have https configuration options by default? https://docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel?view=aspnetcore-3.1#listenoptionsusehttps

PR #233, opened 2019-03-13
Added support for Google buckets storage

bgrainger:
This replaces PR #172 (which hasn't seen active development recently).

- [x] Squash all commits in #172 and resolve conflicts with `master`
- [x] Addressing code review from #172
- [x] Update documentation

### DIFF
--- a/docs/cloud/gcp.md
+++ b/docs/cloud/gcp.md
@@ -1,9 +1,44 @@
-# Running BaGet on the Google Cloud
+# Running BaGet on Google Cloud
 
 !!! warning
     This page is a work in progress!
 
-Sadly, BaGet does not support GCP today. We're open source and accept contributions!
+We're open source and accept contributions!
 [Fork us on GitHub](https://github.com/loic-sharma/BaGet).
 
-For now, please refer to the [Azure documentation](azure).
+## Google Cloud Storage
+
+Packages can be stored in [Google Cloud Storage](https://cloud.google.com/storage/).
+
+### Setup
+
+Follow the instructions in [Using Cloud Storage](https://cloud.google.com/appengine/docs/flexible/dotnet/using-cloud-storage) to:
+
+* Create a bucket
+* Set up a service account and download credentials
+* Set the `GOOGLE_APPLICATION_CREDENTIALS` environment variable to the path to the JSON file you downloaded
+
+### Configuration
+
+Configure BaGet to use GCS by updating the [`appsettings.json`](https://github.com/loic-sharma/BaGet/blob/master/src/BaGet/appsettings.json) file:
+
+```json
+{
+    ...
+
+    "Storage": {
+        "Type": "GoogleCloud",
+        "BucketName": "your-gcs-bucket"
+    },
+
+    ...
+}
+```
+
+## Google Cloud SQL
+
+* TODO
+
+## Google AppEngine
+
+* TODO
--- a/src/BaGet.GCP/Configuration/GoogleCloudStorageOptions.cs
+++ b/src/BaGet.GCP/Configuration/GoogleCloudStorageOptions.cs
@@ -2,7 +2,7 @@ using BaGet.Core.Configuration;
 
 namespace BaGet.GCP.Configuration
 {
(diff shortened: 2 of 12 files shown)


bgrainger:
I'm currently running this code on Google App Engine, and it's working well. (URL available on request.)

docs/cloud/gcp.md
```diff
@@ -1,9 +1,43 @@
...
+
+## Google Cloud SQL
+
+* TODO
+
+## Google AppEngine
+
+* TODO
+
+We're open source and accept contributions! [Fork us on GitHub](https://github.com/loic-sharma/BaGet).
```
loic-sharma: Please move line 43 to somewhere after line 4

loic-sharma:
The code looks great! I'm not very familiar with Google Cloud, is there an emulator I can run on Windows/MacOS that you'd recommend to test these changes?

docs/cloud/gcp.md
loic-sharma: Thanks for taking care of the docs by the way! 👍

bgrainger:
There's no emulator (for Cloud Storage) officially supported by Google that I know of; there are only there four (for different products, and in beta, no less): https://cloud.google.com/sdk/gcloud/reference/beta/emulators/

The relevant issue appears to be https://github.com/googleapis/google-cloud-python/issues/4840.

I was able to test this as my employer is considering using BaGet and we have a GCP account.

src/BaGet.GCP/Configuration/GoogleCloudStorageOptions.cs
```diff
@@ -0,0 +1,9 @@
+using BaGet.Core.Configuration;
+
+namespace BaGet.GCP.Configuration
+{
+    public class GoogleCloudStorageOptions : StorageOptions
+    {
+        public string BucketName { get; set; }
```
loic-sharma: Please mark this as `[Required]`. For example, see: https://github.com/loic-sharma/BaGet/blob/master/src/BaGet.Azure/Configuration/BlobStorageOptions.cs#L7

src/BaGet.GCP/Services/GoogleCloudStorageService.cs
```diff
@@ -0,0 +1,98 @@
...
+
+namespace BaGet.GCP.Services
+{
+    public class GoogleCloudStorageService : IStorageService
+    {
+        private readonly string _bucketName;
+
+        public GoogleCloudStorageService(IOptionsSnapshot<GoogleCloudStorageOptions> options)
+        {
+            _bucketName = options.Value.BucketName;
```
loic-sharma: Please add `if (options == null) throw new ArgumentNullException(nameof(options));` before storing the `_bucketName`

bgrainger: Will do.

BTW, was just copying this existing code; does it need to be fixed? https://github.com/loic-sharma/BaGet/blob/873b133c619e2bc048bbedc5a7fa8007c688eaf8/src/BaGet.AWS/S3StorageService.cs#L22

bgrainger: Maybe never mind; that's `IOptions<>` vs `IOptionsSnapshot<>` and I don't know the difference (specifically in use and nullability).

loic-sharma: Good catch. Yes, both should have null check and both should use `IOptionsSnapshot`. Feel free to fix it in this pull request if you'd like. If not, I can take care of it later

bgrainger: I can push a new commit to add that check.

bgrainger:
All code review was addressed and `null` checks added to `S3StorageService`; this PR is ready for final review.

bgrainger:
@loic-sharma Just checking that you're not waiting on me for anything more before this can be merged?

loic-sharma:
Looks great, thanks @bgrainger!

PR #239, opened 2019-03-18
Private Feeds Part 1

WernerMairl:
this PR contains new code but this code is never called (lowest impact risk for current version)
but i need this code as foundation for the next parts => namespaces to review....

- Credential handling code + tests
- Test refactorings for PackageControllerTests into Fixtures


src/BaGet.Core/Authentication/AuthenticationHeaderValueExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Text;
+
+namespace BaGet.Core.Authentication
+{
+    public class AuthenticationHeaderValueExtensions
+    {
+        public static string GetBase64BasicAuthHeaderString(NetworkCredential credential)
+        {
+            if (credential == null) throw new ArgumentNullException(nameof(credential));
+            return System.Convert.ToBase64String(System.Text.ASCIIEncoding.ASCII.GetBytes(GetBasicAuthHeaderString(credential)));
```
loic-sharma: Please add `using`s for these types.

tests/BaGet.Tests/BasicAuthenticationHeaderBuilderTest.cs
```diff
@@ -0,0 +1,27 @@
+using System.Net;
+using BaGet.Core.Authentication;
+using Xunit;
+namespace BaGet.Tests
+{
+    public class BasicAuthenticationHeaderBuilderTest
+    {
+        [Theory]
+        [InlineData("myUserName","myPassword", "myDomain")]
```
loic-sharma: Please add a space between `,` and `"myPassword"`

src/BaGet.Core/Authentication/BasicAuthenticationHeaderBuilder.cs
```diff
@@ -0,0 +1,13 @@
+using System.Net;
+using System.Net.Http.Headers;
+
+namespace BaGet.Core.Authentication
+{
+    public class BasicAuthenticationHeaderBuilder
+    {
+        public static AuthenticationHeaderValue CreateWith(NetworkCredential credential)
+        {
+            return new System.Net.Http.Headers.AuthenticationHeaderValue(BasicAuthenticationDefaults.AuthenticationScheme, AuthenticationHeaderValueExtensions.GetBase64BasicAuthHeaderString(credential));
```
loic-sharma: Do you need the full namespace here? Can this be simplified to just `AuthenticationHeaderValue`?

src/BaGet.Core/Authentication/AuthenticationHeaderValueExtensions.cs
```diff
@@ -0,0 +1,58 @@
+using System;
+using System.Net;
+using System.Net.Http.Headers;
+using System.Text;
+
+namespace BaGet.Core.Authentication
+{
+    public class AuthenticationHeaderValueExtensions
+    {
+        public static string GetBase64BasicAuthHeaderString(NetworkCredential credential)
```
loic-sharma: From the class name it seems that these should be extension methods on the `NetworkCredential` type. I think this pattern would be pretty nice:

```
NetworkCredential credentials = ...;

credentials.ToBase64BasicAuthHeaderString();
credentials.ToBasicAuthHeaderString();`

File: src/BaGet.Core/Authentication/AuthenticationHeaderValueExtensions.cs [hunk 5]
```diff
@@ -0,0 +1,58 @@
...
+        {
+            if (credential == null) throw new ArgumentNullException(nameof(credential));
+            return System.Convert.ToBase64String(System.Text.ASCIIEncoding.ASCII.GetBytes(GetBasicAuthHeaderString(credential)));
+        }
+
+        public static string GetBasicAuthHeaderString(NetworkCredential credential)
```
loic-sharma: Only `GetBase64BasicAuthHeaderString` is supposed to be called by the authentication pipeline, right? If so, let's make this method private.

PR #240, opened 2019-03-18
Private feeds Part 2

WernerMairl:
Basics for token based auth.

Changes are additive, not changing current code and never called at this moment (calls should coming with part 3 or 4...)

review: namespaces and locations

### DIFF
--- a/src/BaGet.Core.Server/Extensions/NugetBehaviorMiddleware.cs
+++ b/src/BaGet.Core.Server/Extensions/NugetBehaviorMiddleware.cs
@@ -1,0 +1,88 @@
+using System;
+using System.Net.Http.Headers;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+using Microsoft.Net.Http.Headers;
+
+namespace BaGet.Extensions
+{
+    /// <summary>
+    /// Special middleware that translates the non RFC behavior from NuGet.exe into default Bearer RFC behavior before the default JwtBearer Middleware from the AspNetCore Team is called
+    /// This special rewrite of request/response is executed only if the client is the nuget commandline. Other clients (Web-UI) are expected to be implemented RFC conform!
+    /// </summary>
+    public class NugetBehaviorMiddleware
+    {
+        private readonly RequestDelegate NextRequest;
+        private readonly ILogger Logger;
+        private const string BasicAuthenticationScheme = "Basic";
+
+        public NugetBehaviorMiddleware(RequestDelegate next, ILogger<NugetBehaviorMiddleware> logger)
+        {
+            NextRequest = next ?? throw new ArgumentNullException(nameof(next));
+            Logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        private void ModifyRequest(HttpContext context, bool isNuGetClientCall)
+        {
+
+            if (!isNuGetClientCall) return;
+
+            string authorization = context.Request.Headers[HeaderNames.Authorization];
+            // If no authorization header found, nothing to process further
+            if (string.IsNullOrEmpty(authorization))
+            {
+                return;
+            }
+
+            var authHeader = AuthenticationHeaderValue.Parse(authorization);
+            if (string.IsNullOrEmpty(authHeader.Parameter)) return;
+
+            if (authHeader.Scheme == BasicAuthenticationScheme)
+            {
+                var credentialBytes = Convert.FromBase64String(authHeader.Parameter);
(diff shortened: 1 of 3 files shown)


src/BaGet.Core.Server/Extensions/NugetBehaviorMiddleware.cs
```diff
@@ -0,0 +1,88 @@
...
+                }
+                //NuGet.exe has only "Basic" implemented. convert this into "Bearer" directly on the request/response pipeline, then we can use the default jwt implementation from aspnetcore
+                context.Request.Headers[HeaderNames.Authorization] = $"{Microsoft.AspNetCore.Authentication.JwtBearer.JwtBearerDefaults.AuthenticationScheme} {password}";
+                Logger.LogTrace($"Request header field '{HeaderNames.Authorization}' rewritten to '{Microsoft.AspNetCore.Authentication.JwtBearer.JwtBearerDefaults.AuthenticationScheme}'");
+            }
+        }
+
+        private void ModifyResponse(HttpContext context, bool isNuGetClientCall)
+        {
+            if (isNuGetClientCall)
```
loic-sharma: Can we modify all responses regardless of the client here?

src/BaGet.Core.Server/Extensions/NugetBehaviorMiddleware.cs
```diff
@@ -0,0 +1,88 @@
...
+
+namespace BaGet.Extensions
+{
+    /// <summary>
+    /// Special middleware that translates the non RFC behavior from NuGet.exe into default Bearer RFC behavior before the default JwtBearer Middleware from the AspNetCore Team is called
+    /// This special rewrite of request/response is executed only if the client is the nuget commandline. Other clients (Web-UI) are expected to be implemented RFC conform!
+    /// </summary>
+    public class NugetBehaviorMiddleware
+    {
+        private readonly RequestDelegate NextRequest;
```
loic-sharma: Please move the `const` to be before instance members, and please rename private instance fields to `_nextRequest` and `_logger`

loic-sharma:
Hey thanks for figuring all this out, it's looking good so far! :)

Could you explain the purpose of `NugetBehaviorMiddleware`? What's the benefit of translating basic auth to JWT?

src/BaGet.Core.Server/Extensions/NugetBehaviorMiddleware.cs
WernerMairl: that "kills" the UI because the UI is following the "official" Workflow...

what ist your intent/concern ?

WernerMairl:
NuGet Client implementation is NOT following the typical workflow/RFC for Token based security.

NuGet Client has no  (March 2019) support for tokens (a header with AUTH=Bearer).
https://github.com/NuGet/Home/issues/7841

The most common workaround(hack) that i have found is: abuse the basic authentication workflow and send the token using the password field.

on the server side i like to use all the available Token-based staff from aspnetcore team, but this code is not able to handle this "abuse"

my "middleware" does the translation between "NuGet" and "aspnetcore" implementation.

AND
the middleWare should do this Transaltion only in case of a NuGet.client (detected by session ID) to prevent any impact on UI or other implementations that are using token based security in the right way....

src/BaGet.Core.Server/Extensions/NugetBehaviorMiddleware.cs
WernerMairl: done!

loic-sharma:
Ah okay, so this allows you to support AAD without needing a custom NuGet authentication plugin? That’s pretty smart, good work!

WernerMairl:
>support AAD without needing a custom NuGet authentication plugin
not exactly, but in this direction.

Let me explain:
The middleware has a big impact on how we implement the BaGet Backend
It allows us to use all the assemblies/nuget packages for authentication ("Microsoft.AspNetCore.Authentication*") provided by the AspNetCore Team **without** hacks and deep customizing -  we can use them in a default way (!!)



What do you mean with "custom NuGet authentication plugin" ?
I'm assuming do you mean a "NuGetCredentialProvider" that must be used/provided on the client side (Nuget.exe)

We need a plugin for AAD that does the complex UI-Workflow for AAD and other token based security workflows.

BUT: our implementation behaves like all the other NuGet Server implementations (MyGet, Nexus, Azure Artifact) so there is a high probability that we can use "NuGetCredentialProviders" implemented by somewhere else because the implementation must support EXACT the same server behavior.....

With this in mind, the current NuGetCredentialProvider for Azure DevOps/Visual Studio/TFS may work for us, we must only customize the address/instance of AAD.....

PR #257, opened 2019-04-17
[Azure] Add Azure Table Storage database

loic-sharma:
This is an experiment to support massive read/write throughput using Azure Table Storage. I'll be benchmarking this on Azure later. The moonshot goal is to scale this to 12,000 reads per second (~1 billion reads per day), as this is roughly NPM's scale.

* [ ] Perf test
  * [ ] Mass index
  * [ ] Perf test registration leaf
  * [ ] Perf test registration page
  * [ ] Perf test version list
* [ ] Consider adding `IPackageService.FindVersions` API to improve version list perf
* [ ] Consider adding a special row on each package partition that lists that package's versions. It's unclear how to implement this, maintaining this row should not harm write throughput but ideally this row should be strongly consistent


loic-sharma:
Replaced by https://github.com/loic-sharma/BaGet/pull/347

PR #268, opened 2019-05-13
Allow overriding PackageService methods

ismcagdas:
Allow overriding PackageService methods so that a new package service like MyPackageService can be inherited from PackageService instead of implementing IPackageService fully.

### DIFF
--- a/src/BaGet.Core/State/PackageService.cs
+++ b/src/BaGet.Core/State/PackageService.cs
@@ -17,7 +17,7 @@ namespace BaGet.Core.State
             _context = context ?? throw new ArgumentNullException(nameof(context));
         }
 
-        public async Task<PackageAddResult> AddAsync(Package package)
+        public virtual async Task<PackageAddResult> AddAsync(Package package)
         {
             try
             {
@@ -34,7 +34,7 @@ namespace BaGet.Core.State
             }
         }
 
-        public Task<bool> ExistsAsync(string id, NuGetVersion version = null)
+        public virtual Task<bool> ExistsAsync(string id, NuGetVersion version = null)
         {
             var query = _context.Packages.Where(p => p.Id == id);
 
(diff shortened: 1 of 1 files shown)


loic-sharma:
Thanks for opening this pull request! How’re you customizing `PackageService`? I’d like to better understand your scenario better before comitting to a solution.

ismcagdas:
@loic-sharma sure, for now I implement `IPackageService` something like below:

```
public class MyPackageService : IPackageService
{
    // implementation
}
```

Instead of implementing the entire interface, I would like to use it like below:

```
public class MyPackageService : PackageService
{
    public override async Task<IReadOnlyList<Package>> FindAsync(string id, bool includeUnlisted = false)
    {
        // My custom implementation
    }
}
```

Then, use MyPackageService like below:

`services.AddTransient<IPackageService, MyPackageService>();`

loic-sharma:
Could you instead do something like:

```csharp
public class MyPackageService : IPackageService
{
    private readonly PackageService _defaultPackageService;

    public MyPackageService(PackageService defaultPackageService)
    {
        _defaultPackageService = defaultPackageService ?? throw new ArgumentNullException(nameof(defaultPackageService));
    }

    public Task<PackageAddResult> AddAsync(Package package) => _defaultPackageService.AddAsync(package);
    public Task<bool> Exists(string id, NuGetVersion version = null) => _defaultPackageService.Exists(id, version);

    // ...

    public override async Task<IReadOnlyList<Package>> FindAsync(string id, bool includeUnlisted = false)
    {
        // My custom implementation
    }
}
```

You'll need to tweak the dependency injection a little:

```csharp
services.AddTransient<PackageService>();
services.AddTransient<IPackageService, MyPackageService>();
```

I acknowledge that this isn't as convenient as overriding methods, however, I'd prefer this over setting a precedent of virtual methods for extensibility. Let me know what you think!

ismcagdas:
@loic-sharma no problem, I will use it like that, thank you.

loic-sharma:
Please let me know if you run into any issues or have any additional feedback. Cheers!

PR #269, opened 2019-05-13
Allow a way to filter packages in DatabaseSearchService

ismcagdas:
In some cases, users might want to filter package search results according to a custom logic. Introducing a virtual method like FilterPackages in DatabaseSearchService.cs can help to do that.


loic-sharma:
What kind of custom logic are you adding to filter packages?

ismcagdas:
We have another table on the database which contains customers and license expiration date for customers. We want to show only packages published before the license expiration date for the customers.

loic-sharma:
I see. Could you wrap the `DatabaseSearchService` similar to the approach described [here](https://github.com/loic-sharma/BaGet/pull/268#issuecomment-492495455)? This seems like a cleaner approach.

ismcagdas:
@loic-sharma no problem, I will use it like that, thank you 😄

Issue #266, opened 2019-05-07
Can BaGet authenticate with S3 via IAM roles?


richarddowner:
### Is your feature request related to a problem? Please describe.
I do not want to use AccessKeys within AWS. I want to use assumed roles / instance profiles instead. 

### Describe the solution you'd like
The docs here: https://loic-sharma.github.io/BaGet/cloud/aws/ describe AccessKey and SecretKey.
A quick fix would be if the AccessKey and SecretKey are not specified, fall back to using the InstanceProfile keys: AccessKey SecretKey and Token in the S3 client credentials for the requests. These can be obtained via the FallbackCredentialsFactory.GetCredentials() which returns a AWSCredentials object containing the AccessKey, SecureSecretKey and Token.

lucasbayma:
The workaround that I've found was to create a "Baget" dummy account in AWS, with just the permissions to update s3 and use the accesKey and secretKey from that account.

loic-sharma:
Thanks for the suggestion! What do you think of adding a `UseInstanceProfile` config like:

```json
{
   // ...

    "Storage": {
        "Type": "AwsS3",
        "Region": "us-west-1",
        "Bucket": "foo",
        "UseInstanceProfile": true
    },

   // ...
}
```

If `UseInstanceProfile` is true, BaGet uses `FallbackCredentialsFactory.GetCredentials()`. If `UseInstanceProfile` is false, BaGet will require an access key and secret key. Does this seem like a reasonable approach?

richarddowner:
Yea this sounds perfect. 
If the container uses the FallbackCredentialsFactory all you will need to do is make sure the task definition has a taskrole with access to the S3 bucket.

richarddowner:
Just got this working locally with the FallbackCredentialsFactory, however I am concerned the temp credentials may expire. I know there is a way to use a client type that auto-refreshes the temp creds. Might need to use that instead. I will figure it out and send a PR eventually. 

Also, how does one update the documentation? Is it part of the project?

Cheers

richarddowner:
I have done the commit but will need to test it within AWS. Will send PR when ready...
https://github.com/richarddowner/BaGet/commit/61a999e46f129fcaa220e2ce702bca4d6e73bbea

PR #277, opened 2019-06-03
Two additional ways of access in AWS S3

richarddowner:

 * Added UseInstanceProfile config option, for using fallback credentials (instance profile) in AWS
 * Added AssumeRoleArn config option, for assuming cross account role in AWS
 * Todo: Update documentation, not sure where to do this?

Addresses https://github.com/loic-sharma/BaGet/issues/266

### DIFF
--- a/src/BaGet.AWS/Configuration/S3StorageOptions.cs
+++ b/src/BaGet.AWS/Configuration/S3StorageOptions.cs
@@ -1,4 +1,4 @@
-﻿using System.ComponentModel.DataAnnotations;
+using System.ComponentModel.DataAnnotations;
 using BaGet.Core.Validation;
 
 namespace BaGet.AWS.Configuration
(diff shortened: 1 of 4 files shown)


loic-sharma:
Thank you for doing this work, this looks great! I'll send a pull request to update the documentation, and I'll tag you on it so that you can review it.

altso:
@richarddowner @loic-sharma I just tried to use in `AssumeRoleArn` and got this exception:

```
System.InvalidOperationException: Assembly AWSSDK.SecurityToken could not be found or loaded.
This assembly must be available at runtime to use Amazon.Runtime.AssumeRoleAWSCredentials.
```

It looks similar to https://github.com/aws/aws-lambda-dotnet/issues/223 and the resolution there was to add `AWSSDK.SecurityToken` as a dependency.

I can create a pull request with this change if you think it's the way to go.

loic-sharma:
@altso Thanks for reporting this! Your fix sounds good to me, I'd be happy to accept a pull request :)

altso:
@loic-sharma created #460

PR #280, opened 2019-06-09
Fix .NET Framework builds on macOS/Linux

loic-sharma:


src/BaGet/appsettings.json
```diff
@@ -18,7 +18,7 @@
   },
 
   "Mirror": {
-    "Enabled": false,
```
nvoskuilen: Consider reverting this commit, see #300

src/BaGet/appsettings.json
```diff
@@ -18,7 +18,7 @@
...
+    "Enabled": true,
```
nvoskuilen: Consider reverting this commit, see #300

loic-sharma: Thanks for reporting this, I've reverted it here https://github.com/loic-sharma/BaGet/commit/79ab701ab16b12f3ee951e7d43ffc90a85ad3f6a

PR #287, opened 2019-06-24
Provide the option to get DB connection string from ENV variable.

richarddowner:

 * Add the option to specify connection string via environment variable
 * Allows docker containers to resolve secrets at run time rather than hard code them into connection strings withing the application. 
* This is a quick solution, a better solution would be to provide a generic platform agnostic way of accessing secrets at run time then building the connection string within the application itself.  

(Not sure why my old commits are shown here as well, those are already merged)

### DIFF
--- a/src/BaGet.AWS/Configuration/S3StorageOptions.cs
+++ b/src/BaGet.AWS/Configuration/S3StorageOptions.cs
@@ -1,4 +1,4 @@
-﻿using System.ComponentModel.DataAnnotations;
+using System.ComponentModel.DataAnnotations;
 using BaGet.Core.Validation;
 
 namespace BaGet.AWS.Configuration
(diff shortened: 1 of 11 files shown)


szwenni:
Hi,
I have seen your pull request.
From my testings you could provide the Environment Variable `Database__ConnectionString`.
See https://loic-sharma.github.io/BaGet/quickstart/docker/. As these are just Environment Variables you could also add them to the Docker run command with the -e option.

Hope this helps you.

loic-sharma:
I missed this pull request somehow, sorry about that! szwenni is correct, you can find more information about using environment variables here: https://loic-sharma.github.io/BaGet/configuration/#environment-variables

Do you mind if I close this pull request?

loic-sharma:
Closing as per szwenni's comment: https://github.com/loic-sharma/BaGet/pull/287#issuecomment-545422001

Issue #276, opened 2019-05-29
Cannot push nuget packages to BaGet behind an  IIS Reverse Proxy


marcBalta:
### Describe the bug
I'm using BaGet inside a windows docker image. IIS 10 is used as a reverse proxy. 
This is the web.config file used to proxy queries with the pattern domain.de/nuget/ to the BaGet server which is running in a Windows Docher container on another machine with the IP 192.168.178.2.

'''
		 <rule name="nuget" stopProcessing="true">
                    <match url="nuget/(.*)" />
                    <action type="Rewrite" url="http://192.168.178.2:5555/{R:1}" />
                    <serverVariables>
                        <set name="HTTP_FORWARDED" value="for={REMOTE_ADDR};by={LOCAL_ADDR};host=&quot;{HTTP_HOST}&quot;;proto=&quot;https&quot;" />
			<set name="HTTP_X_FORWARDED_HOST" value="{HTTP_HOST}" />
			<set name="HTTP_X_FORWARDED_SCHEMA" value="https" />
			<set name="HTTP_X_FORWARDED_PROTO" value="https" />
                    </serverVariables>
                </rule>

This is my baget.env :
'''				
ApiKey=NUGET-SERVER-API-KEY
PathBase=/nuget
Storage__Type=FileSystem
Storage__Path=c:\packages
Database__Type=Sqlite
Database__ConnectionString=Data Source=C:\packages\baget.db
Search__Type=Database
'''

When Im running however the nuget push command it uses the internal IP address (the url of docker container) and not the one of the proxy.
'''
$ dotnet nuget push  -s https://ge-apps-dev.intranet.mckinsey.com/nuget/v3/index.json -k NUGET-SERVER-API-KEY McK.Toolkit.Core.Asp.1.0.0.nupkg
info : Pushing McK.Toolkit.Core.Asp.1.0.0.nupkg to 'https://192.168.178.2:5555/api/v2/package'...
info :   PUT https://192.168.178.2:5555/api/v2/package/
info : An error was encountered when fetching 'PUT https://192.168.178.2:5555/api/v2/package/'. The request will now be retried.
info : A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond
info :   PUT https://192.168.178.2:5555/api/v2/package/
info : An error was encountered when fetching 'PUT https://192.168.178.2:5555/api/v2/package/'. The request will now be retried.
[...]

PR #388, opened 2019-10-17
adding basepath support

szwenni:

 * Adding full PathBase support
 * Adding dynamic PathBase replacing in react scripts on StartUp of server

Addresses https://github.com/loic-sharma/BaGet/issues/276

### DIFF
--- a/src/BaGet.Core.Server/Configuration/ConfigureForwardedHeadersOptions.cs
+++ b/src/BaGet.Core.Server/Configuration/ConfigureForwardedHeadersOptions.cs
@@ -1,15 +1,31 @@
+using BaGet.Core;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.HttpOverrides;
 using Microsoft.Extensions.Options;
+using System;
+using System.Collections.Generic;
 
 namespace BaGet.Configuration
 {
     public class ConfigureForwardedHeadersOptions : IConfigureOptions<ForwardedHeadersOptions>
     {
+        private BaGetOptions bagetOptions;
+        public ConfigureForwardedHeadersOptions(IOptions<BaGetOptions> bagetOptions)
+        {
+            this.bagetOptions = bagetOptions.Value;
+        }
+
         public void Configure(ForwardedHeadersOptions options)
         {
-            options.ForwardedHeaders = ForwardedHeaders.XForwardedFor | ForwardedHeaders.XForwardedProto;
-
+            if (!string.IsNullOrEmpty(bagetOptions.HostUrl))
+            {
+                Console.WriteLine("Host Url: " + bagetOptions.HostUrl);
+                options.ForwardedHeaders = ForwardedHeaders.XForwardedFor | ForwardedHeaders.XForwardedProto | ForwardedHeaders.XForwardedHost;
+                options.AllowedHosts = new List<string>() { bagetOptions.HostUrl};
+            }else
+            {
+                options.ForwardedHeaders = ForwardedHeaders.XForwardedFor | ForwardedHeaders.XForwardedProto;
+            }
             // Do not restrict to local network/proxy
             options.KnownNetworks.Clear();
             options.KnownProxies.Clear();
--- a/src/BaGet.Core/Configuration/BaGetOptions.cs
+++ b/src/BaGet.Core/Configuration/BaGetOptions.cs
@@ -17,6 +17,10 @@ namespace BaGet.Core
         public string PathBase { get; set; }
 
         /// <summary>
+        /// The application Host without port number. If not set all hosts will be allowed. See https://github.com/aspnet/Announcements/issues/295
+        /// </summary>
+        public string HostUrl { get; set; }
+        /// <summary>
         /// If enabled, the database will be updated at app startup by running
         /// Entity Framework migrations. This is not recommended in production.
         /// </summary>
--- a/src/BaGet.UI/src/index.tsx
+++ b/src/BaGet.UI/src/index.tsx
@@ -1,13 +1,14 @@
 import { initializeIcons } from '@uifabric/icons';
 import 'bootstrap/dist/css/bootstrap.css';
 import { BrowserRouter as Router, Route } from 'react-router-dom';
-
+import { config } from './config';
 import * as React from 'react';
 import * as ReactDOM from 'react-dom';
 import App from './App';
 
 import DisplayPackage from './DisplayPackage/DisplayPackage';
 import Upload from './Upload';
+import createBrowserHistory from 'history/createBrowserHistory';
 
 import './index.css';
 
@@ -16,7 +17,7 @@ import './index.css';
 initializeIcons();
 
 ReactDOM.render(
-  <Router>
+  <Router basename={config.baseUrl}>
     <App>
       <Route path="/packages/:id/:version?" component={DisplayPackage} />
 
--- a/src/BaGet/Startup.cs
+++ b/src/BaGet/Startup.cs
@@ -54,7 +54,17 @@ namespace BaGet
                 }
             }
 
-            app.UsePathBase(options.PathBase);
+            //only rewrite the path values of react files in production
+            //in dev mode please change config.tsx and the package.json homepage value before building react
+            if (env.IsDevelopment())
+            {
+                app.UsePathBase(options.PathBase);
+            }
+            else
+            {
+                app.UsePathBase(options, "BaGet.UI/build");
+            }
+            
             app.UseForwardedHeaders();
             app.UseSpaStaticFiles();
 
--- a/src/BaGet.Core.Server/Extensions/IApplicationBuilderExtensions.cs
+++ b/src/BaGet.Core.Server/Extensions/IApplicationBuilderExtensions.cs
@@ -1,4 +1,5 @@
 using System;
+using System.IO;
 using Microsoft.AspNetCore.Builder;
 
 namespace BaGet.Core.Server.Extensions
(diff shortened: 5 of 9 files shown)


src/BaGet.Core.Server/Configuration/ConfigureForwardedHeadersOptions.cs
```diff
@@ -1,15 +1,31 @@
+using BaGet.Core;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.HttpOverrides;
 using Microsoft.Extensions.Options;
+using System;
```
loic-sharma: Please place put `using System*` first. Visual Studio can do this automatically if you enable the option `Place 'System' directives when sorting usings` under `Text Editor` > `C#` > `Advanced`.

src/BaGet.Core.Server/Configuration/ConfigureForwardedHeadersOptions.cs
```diff
@@ -1,15 +1,31 @@
...
+using System.Collections.Generic;
 
 namespace BaGet.Configuration
 {
     public class ConfigureForwardedHeadersOptions : IConfigureOptions<ForwardedHeadersOptions>
     {
-        public void Configure(ForwardedHeadersOptions options)
+        private BaGetOptions bagetOptions;
```
loic-sharma: Please add the `readonly` field modifier to ensure no one accidentally modifies this value.
Also, please keep the options wrapped in a `IOptionsSnapshot` to support automatic configuration reload.

```suggestion
        private readonly IOptionsSnapshot<BaGetOptions> _options;
```

src/BaGet.Core.Server/Configuration/ConfigureForwardedHeadersOptions.cs
```diff
@@ -1,15 +1,31 @@
...
+        }
 
+        public void Configure(ForwardedHeadersOptions options)
+        {
+            if (!string.IsNullOrEmpty(bagetOptions.HostUrl))
+            {
+                Console.WriteLine("Host Url: " + bagetOptions.HostUrl);
+                options.ForwardedHeaders = ForwardedHeaders.XForwardedFor | ForwardedHeaders.XForwardedProto | ForwardedHeaders.XForwardedHost;
+                options.AllowedHosts = new List<string>() { bagetOptions.HostUrl};
+            }else
```
loic-sharma: Please add new lines in between blocks:

```csharp
if (...)
{
  ...
}
else
{
  ..
}

// Do not restrict to local network/proxy
```

src/BaGet.Core.Server/Configuration/ConfigureForwardedHeadersOptions.cs
```diff
@@ -1,15 +1,31 @@
...
         {
-            options.ForwardedHeaders = ForwardedHeaders.XForwardedFor | ForwardedHeaders.XForwardedProto;
+            this.bagetOptions = bagetOptions.Value;
+        }
 
+        public void Configure(ForwardedHeadersOptions options)
+        {
+            if (!string.IsNullOrEmpty(bagetOptions.HostUrl))
+            {
+                Console.WriteLine("Host Url: " + bagetOptions.HostUrl);
```
loic-sharma: Can we remove this? Or do you think this will be useful for customers? If the latter, please log using an `ILogger<ConfigureForwardedHeaderOptions>` as not everyone has access to stdout to browse logs

src/BaGet.Core/Configuration/BaGetOptions.cs
```diff
@@ -16,6 +16,10 @@ public class BaGetOptions
         /// </summary>
         public string PathBase { get; set; }
 
+        /// <summary>
+        /// The application Host without port number. If not set all hosts will be allowed. See https://github.com/aspnet/Announcements/issues/295
```
loic-sharma: Nice link! 👍

src/BaGet.UI/src/index.tsx
```diff
@@ -1,13 +1,14 @@
...
 import { BrowserRouter as Router, Route } from 'react-router-dom';
-
+import { config } from './config';
 import * as React from 'react';
 import * as ReactDOM from 'react-dom';
 import App from './App';
 
 import DisplayPackage from './DisplayPackage/DisplayPackage';
 import Upload from './Upload';
+import createBrowserHistory from 'history/createBrowserHistory';
```
loic-sharma: What's this for?

loic-sharma:
Hi, this is awesome, thank you for opening this change! I wasn't sure how to share configs between the backend and the frontend, this is a really neat solution! I'm wondering, how did you choose to override the frontend configs when the app middleware is created in `UsePathBase`?

I wonder if instead we should upgrade BaGet's [`AddSpaStaticFiles`](https://github.com/loic-sharma/BaGet/blob/be8caa65c52ab9ecad27509b3116c65601827ad2/src/BaGet/Startup.cs#L29). `AddSpaStaticFiles` registers an [`ISpaStaticFileProvider`](https://github.com/aspnet/AspNetCore/blob/master/src/Middleware/SpaServices.Extensions/src/StaticFiles/DefaultSpaStaticFileProvider.cs#L46), which exposes an [`IFileProvider`](https://github.com/aspnet/AspNetCore/blob/master/src/Middleware/SpaServices.Extensions/src/StaticFiles/DefaultSpaStaticFileProvider.cs#L50). We could create our own `ISpaStaticFileProvider` that has a custom `IFileProvider` that replaces configurations lazily. What do you think of that approach? Let me know if you need help with that!

Again, thank you for opening this change. You did a phenomenal job!

src/BaGet.Core.Server/Configuration/ConfigureForwardedHeadersOptions.cs
szwenni: Okay i will change that in next commit. I wasn't aware of that option.

src/BaGet.Core.Server/Configuration/ConfigureForwardedHeadersOptions.cs
szwenni: I will update this also in next commit. I wasn't aware of the IOptionsSnapshot thing if I'm honest, but it makes sense.

src/BaGet.Core.Server/Configuration/ConfigureForwardedHeadersOptions.cs
szwenni: Yes

src/BaGet.Core.Server/Configuration/ConfigureForwardedHeadersOptions.cs
szwenni: Mabe leaving it hast Debug message should be fine. If somehow problems occur you could enable debug and this might help.

I will take the logger framework. I wanted to change this before pull request but forgot it.

src/BaGet.Core/Configuration/BaGetOptions.cs
szwenni: Found it in your comments and it makes sense to explain why this is required.

src/BaGet.UI/src/index.tsx
szwenni: Forgot to remove it.
I tried adding the history to react router and setting the base path via the history.
But dom router does not support history so I had to use the basename option.

szwenni:
Hi, thnk you very much.
Using `UseBasePath` was the most straight foreward approach for that.
I didn't dig to much into this but your approach seems to be a nicer way to do this.

So basically from the approach I took we have to always place our configuration in the UI files because I choosed a placeholder which will be replaced. So I would suggest to always have the files in-memory. From performance side there should be little to no impact.
I think implementing shouldn't be that hard. I will try to make the required changes.

src/BaGet.Core.Server/Configuration/ConfigureForwardedHeadersOptions.cs
szwenni: I couln't get it to work with `IOptionsSnapshot` because BaGetOptions and ConfigureForwardedHeadersOptions are in different scopes. Maybe you can get it to work.

szwenni:
So
I made the required changes to get it to work with `ISpaStaticFileProvider`.

Take a look at it and tell me your thoughts. I think its a quite neat way to have this in-memory but still with reasonable performance.

szwenni:
So with the latest commit tests are working again.

Still code cleanup needs to be done.

Also we should consinder to write some tests for the StaticFileProviding.

If you have time you can have a look at the implementation and tell me your findings.

src/BaGet.Core.Server/Configuration/ConfigureForwardedHeadersOptions.cs
```diff
@@ -1,15 +1,36 @@
...
-            options.ForwardedHeaders = ForwardedHeaders.XForwardedFor | ForwardedHeaders.XForwardedProto;
+            _options = bagetOptions;
+            _logger = logger;
+        }
 
+        public void Configure(ForwardedHeadersOptions options)
+        {
+            if (!string.IsNullOrEmpty(_options.Value.HostUrl))
+            {
+                _logger.LogDebug("Host Url: " + _options.Value.HostUrl);
```
loic-sharma: It's better to let the logging framework format the log message. This improves performance if the log level is set higher than debug (as the log will just be dropped). Example:

```suggestion
                _logger.LogDebug("Allowing host URL {HostUrl}", _options.Value.HostUrl);
```

src/BaGet/Startup.cs
```diff
@@ -75,12 +80,15 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
...
             {
                 spa.Options.SourcePath = "../BaGet.UI";
-
+                var spaStaticFilesService = app.ApplicationServices.GetService<ISpaStaticFileProvider>();
+                spa.Options.DefaultPageStaticFileOptions = new StaticFileOptions();
+                spa.Options.DefaultPageStaticFileOptions.FileProvider = spaStaticFilesService.FileProvider;
                 if (env.IsDevelopment())
                 {
-                    spa.UseReactDevelopmentServer(npmScript: "start");
+                    //spa.UseReactDevelopmentServer(npmScript: "start");
```
loic-sharma: Do we need to support a custom host name while in development mode? That seems like a weird use case to me. Would it be better if we didn't support a custom host name while in development mode? The React development server is very useful :)

szwenni: Yeah you are right. I don't know enough about the react environment. But maybe it is possible to have like dev environment where homepage is set to `""` and a prodution environment where homepage is set to `__BAGET_PATH_BASE_PLACEHOLDER__`. So that npm (react) build setups the runtime and the index.html relative paths correctly.

I will try to find a appropiate solution for that.

szwenni:
The Placeholder for path base will be now set only in production mode of npm (`npm run build`), if you use `npm start` no relative path will be appended to the start.

daniel-lerch:
Which changes have to be made for this PR to get merged?
I have to use _BaGet_ with a custom `PathBase` and implemented it for the backend in #114.
Please let me know if there is anything that I can do to complete this feature.

szwenni:
I don't know you need to ask @loic-sharma

LaplancheMaxime:
Do you have any idea when the basepath feature will be available?

LaplancheMaxime:
Hi, I need this feature. Do you need help completing it?
Have a nice day

daniel-lerch:
I just took a look into the code and there is a lot of cleanup work that needs to be done. Some classes look more like Java code than C# and ASP.NET has a lot of patterns which the rest of BaGet is following.

Instead of writing dozens of review comments it would be easier for me to fix stylistic mistakes directly.
@szwenni Do you want to give me push access to your fork or shall I open a new pull request?

Issue #403, opened 2019-10-23
Remove interfaces with only one implementation


loic-sharma:
Like `IPackageContent` and `IPackageMetadata`. Make their methods virtual to allow mocking.

loic-sharma:
I had a change of heart on this idea once @michal880 did this work. See https://github.com/loic-sharma/BaGet/pull/409#issuecomment-586656656:

> I like the idea of reducing the number of types in BaGet, however, BaGet is designed to be modular. Interfaces are a clean way to declare "this is what you need to implement", and I think that's more important than reducing the number of types at this point.

PR #409, opened 2019-10-26
Removed interfaces with only one implementation (#403)

michal880:

 Remove interfaces with only one implementation
 Adjusted DI for above
 Fixed their docs
 Fixed some minor "TODO"s

Addresses https://github.com/loic-sharma/BaGet/issues/403


src/BaGet.Core/Indexing/SymbolIndexingService.cs
```diff
@@ -110,10 +141,9 @@ public async Task<SymbolIndexingResult> IndexAsync(Stream stream, CancellationTo
 
                 return files.Where(p => Path.GetExtension(p) == ".pdb").ToList();
             }
-            catch (Exception)
+            catch (Exception e)
             {
-                // TODO: ValidatePackageEntries throws PackagingException.
-                // Add better logging.
+                _logger.LogError(e, "Unable to get symbol package Pdb paths due to exception");
```
loic-sharma: Nice log statement!

Could you re-add the `// TODO: ValidatePackageEntries throws PackagingException.` comment? I *think* we should return `SymbolIndexingResult.InvalidSymbolPackage` in that case, but it's been a while since I've looked at this path.

src/BaGet.Core/Metadata/PackageMetadataService.cs
```diff
@@ -24,8 +28,14 @@ public class DefaultPackageMetadataService : IPackageMetadataService
             _packages = packages ?? throw new ArgumentNullException(nameof(packages));
             _builder = builder ?? throw new ArgumentNullException(nameof(builder));
         }
-
-        public async Task<BaGetRegistrationIndexResponse> GetRegistrationIndexOrNullAsync(
+        /// <summary>
```
loic-sharma: Please add a new line above this one

src/BaGet.Core/Metadata/PackageMetadataService.cs
```diff
@@ -40,16 +50,22 @@ public class DefaultPackageMetadataService : IPackageMetadataService
                     packageId,
                     packages));
         }
-
-        public async Task<BaGetRegistrationLeafResponse> GetRegistrationLeafOrNullAsync(
-            string id,
-            NuGetVersion version,
+        /// <summary>
```
loic-sharma: Please add a new line above this one

src/BaGet.Core/ServiceIndex/BaGetServiceIndex.cs
```diff
@@ -28,8 +33,12 @@ private IEnumerable<ServiceIndexItem> BuildResource(string name, string url, par
                 };
             }
         }
-
-        public Task<ServiceIndexResponse> GetAsync(CancellationToken cancellationToken = default)
+        /// <summary>
```
loic-sharma: Please add a new line above this line.

src/BaGet.Core/ServiceIndex/BaGetServiceIndex.cs
```diff
@@ -6,10 +6,15 @@
...
+    /// <summary>
+    /// The NuGet Service Index service, used to discover other resources.
+    /// 
+    /// See https://docs.microsoft.com/en-us/nuget/api/service-index
+    /// </summary>
+    public class BaGetServiceIndex
     {
         private readonly IUrlGenerator _url;
-
+
```
loic-sharma: Please remove this whitespace

src/BaGet.Core/ServiceIndex/BaGetServiceIndex.cs
```diff
@@ -6,10 +6,15 @@
 
 namespace BaGet.Core
 {
-    public class BaGetServiceIndex : IServiceIndexService
+    /// <summary>
+    /// The NuGet Service Index service, used to discover other resources.
+    /// 
+    /// See https://docs.microsoft.com/en-us/nuget/api/service-index
+    /// </summary>
+    public class BaGetServiceIndex
```
loic-sharma: I think we can rename this to just `ServiceIndex` now? I believe I called this type `BaGetServiceIndex` because the `ServiceIndexClient` type used to be called `ServiceIndex`

src/BaGet.Core/Storage/PackageStorageService.cs
```diff
@@ -113,23 +127,48 @@ public class PackageStorageService : IPackageStorageService
                 lowercasedId,
                 lowercasedNormalizedVersion);
         }
-
-        public async Task<Stream> GetPackageStreamAsync(string id, NuGetVersion version, CancellationToken cancellationToken)
+        /// <summary>
```
loic-sharma: Please add a new line in between all `}` and comments in this file

src/BaGet.Core/ServiceIndex/BaGetServiceIndex.cs
michal880: please recheck, I think it aligns with current convention

src/BaGet.Core/ServiceIndex/BaGetServiceIndex.cs
michal880: sure

src/BaGet.Core/Indexing/SymbolIndexingService.cs
```diff
@@ -158,7 +188,7 @@ bool IsValidSymbolFileInfo(FileInfo file)
                         signature = id.Guid.ToString("N").ToUpperInvariant();
                     }
 
-                    var fileName = Path.GetFileName(pdbPath).ToLowerInvariant();
+                    var fileName = Path.GetFileName(pdbPath)?.ToLowerInvariant();
```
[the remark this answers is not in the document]
michal880: I added this because we have no assertion in this method that string parameter won't be null - makes the method little more defensive. I can remove it if you think it has negative impact.

michal880:
@loic-sharma how would you suggest to handle current CI issues?

loic-sharma:
@michal880 You can add a protected parameterless constructor for mocking. Example: https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/storage/Azure.Storage.Blobs/src/BlobServiceClient.cs#L106

michal880:
should be fine now

michal880:
@loic-sharma can you check if it's ready for merge?

loic-sharma:
Hey @michal880 thanks for opening this. I've been dragging my feet on this because I had a change of heart on this idea once I saw the implementation. I like the idea of reducing the number of types in BaGet, however, BaGet is designed to be modular. Interfaces are a clean way to declare "this is what you need to implement", and I think that's more important than reducing the number of types at this point. I'm sorry for changing my mind so late, especially after you've done such good work. Thanks again for this great work, it helped me make a more informed decision!

Issue #417, opened 2019-11-08
Package search ignores prerelease and other query parameters


tiksn:
### Package search ignores prerelease and other query parameters
When using database (Sqlite in our case) search for latest package returns prerelease packages even when prerelease is explicitly passed as false.

### To Reproduce
1. Using this version of BaGet 'v0.2.0-preview1' and '0.1.77-prerelease' (tested only on those 2)
2. Run 'dotnet run' in folder '.\src\\Buget'
3. Push 2 packages, release and prerelease where prepelease version is greater (like 1.2.3 and 1.2.4-alpha)
4. Run PowerShell Core command `((Invoke-WebRequest -Uri "http://localhost:50561/v3/search?semVerLevel=2.0.0&q=webclient&prerelease=false") | ConvertFrom-Json).Data` where 'webclient' is your package name.
4. See output

### Expected behavior
One record with stable release version

### Actual behavior
One record with prerelease version

PR #418, opened 2019-11-08
Apply search filters on database search results

tiksn:

 * separated query composition into local function
 * calling that function for initial query creation and after query changes

Addresses https://github.com/loic-sharma/BaGet/issues/417

### DIFF
--- a/src/BaGet.Core/Search/DatabaseSearchService.cs
+++ b/src/BaGet.Core/Search/DatabaseSearchService.cs
@@ -178,7 +178,7 @@ namespace BaGet.Core
             var frameworks = GetCompatibleFrameworksOrNull(framework);
             var search = (IQueryable<Package>)_context.Packages.Where(p => p.Listed);
 
-            IQueryable<Package> ComposePackageQuery(IQueryable<Package> packageQuery)
+            IQueryable<Package> AddSearchFilters(IQueryable<Package> packageQuery)
             {
                 if (!string.IsNullOrEmpty(query))
                 {
--- a/src/BaGet.Core/Search/DatabaseSearchService.cs
+++ b/src/BaGet.Core/Search/DatabaseSearchService.cs
@@ -206,6 +206,8 @@ namespace BaGet.Core
                     packageQuery = packageQuery.Where(p => p.TargetFrameworks.Any(f => frameworks.Contains(f.Moniker)));
                 }
 
+                packageQuery = packageQuery.Where(p => p.Listed);
+
                 return packageQuery;
             }
 
@@ -236,7 +238,7 @@ namespace BaGet.Core
 
             search = AddSearchFilters(search);
 
-            var results = await search.Where(p => p.Listed).ToListAsync(cancellationToken);
+            var results = await search.ToListAsync(cancellationToken);
 
             return results.GroupBy(p => p.Id).ToList();
         }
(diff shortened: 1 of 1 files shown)


src/BaGet.Core/Search/DatabaseSearchService.cs
```diff
@@ -178,32 +178,39 @@ public Task IndexAsync(Package package, CancellationToken cancellationToken)
             var frameworks = GetCompatibleFrameworksOrNull(framework);
             var search = (IQueryable<Package>)_context.Packages.Where(p => p.Listed);
 
-            if (!string.IsNullOrEmpty(query))
+            IQueryable<Package> ComposePackageQuery(IQueryable<Package> packageQuery)
```
loic-sharma: What do you think of naming this `AddSearchFilters`?

```suggestion
            IQueryable<Package> AddSearchFilters(IQueryable<Package> packageQuery)
```

loic-sharma: Nice idea using a local function, it's a very clean solution 👍

src/BaGet.Core/Search/DatabaseSearchService.cs
```diff
@@ -178,32 +178,39 @@ public Task IndexAsync(Package package, CancellationToken cancellationToken)
...
             {
```
loic-sharma: We could add `packageQuery = packageQuery.Where(p => p.Listed);` here.

This lets you simplify line 179 to:

```csharp
var search = (IQueryable<Package>)_context.Packages;
```

And lines 237-239 to:

```csharp
var results = await AddSearchFilters(search).ToListAsync(cancellationToken);
```

src/BaGet.Core/Search/DatabaseSearchService.cs
```diff
@@ -178,32 +178,39 @@ public Task IndexAsync(Package package, CancellationToken cancellationToken)
...
-                query = query.ToLower();
-                search = search.Where(p => p.Id.ToLower().Contains(query));
-            }
+                if (!string.IsNullOrEmpty(query))
+                {
+                    query = query.ToLower();
+                    packageQuery = packageQuery.Where(p => p.Id.ToLower().Contains(query));
```
loic-sharma: We only need to filter on `query` once, could you move lines 183-187 out of `ComposePackageQuery` (or `AddSearchFilters`)? I would make line 179 something like:

```csharp
var search = AddSearchFilters(_context.Packages);
if (!string.IsNullOrEmpty(query))
{
  query = query.ToLower();
  search = search.Where(p => p.Id.ToLower().Contains(query));
}

var packageIds = ...
```

loic-sharma:
Thank you for opening this! I left some minor suggestions, I'll merge this in once you've replied 😊

src/BaGet.Core/Search/DatabaseSearchService.cs
tiksn: I would like to change/update `search` before using it. Maybe someone later will add some more lines of code, and I would like to have this filtering in that search already. Please let me know if you agree with it.

loic-sharma:
Your changes look good. Before I merge this in, I will run the following test plan:

1. Create and upload packages:
    1. Create a package with versions:
        * 4.0.0-prerelease+semver2
        * 3.0.0+semver2
        * 2.0.0-prerelease
        * 1.0.0
    1. Create a package with a single prerelease and semver2 version
    1. Create a package with a single prerelease version
    1. Create a package with a single semver2 version
    1. Create a package with a single "normal" version
1. For each database, check that expected package shows up for combinations of prerelease and semver2 filter:
    * [ ] SQLite
    * [ ] PostgreSQL
    * [ ] MySQL
    * [ ] SQL Server

loic-sharma:
Thank you for fixing this!

Issue #408, opened 2019-10-24
Improve error handling on misconfigurations


loic-sharma:
### Is your feature request related to a problem? Please describe.
BaGet will throw an exception if it encounters invalid configurations. This is not very helpful.

### Describe the solution you'd like
Add helpful error messages that link to the documentation.

MarkZither:
I think this moves all the validation to startup rather than when a section is first used, still working through all config options.

PR #449, opened 2020-01-23
[WIP] Improve error handling on misconfigurations

MarkZither:
I have started looking at how this can be improved. One option is to use a StartupFilter ([see Andrew Lock's blog](https://andrewlock.net/adding-validation-to-strongly-typed-configuration-objects-in-asp-net-core/)) this builds on the existing config validation code and just runs all the validation at startup instead of when a config section is first used.

There are still some options which are used before the validation happens and enums which are not valid are not handled.

Addresses https://github.com/loic-sharma/BaGet/issues/408


loic-sharma:
Hi, thanks for opening this! ASP.NET Core 3 added [option validation](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-3.1#options-validation
), which looks promising. I think BaGet should move off its custom options validation and onto ASP.NET Core's. Once we've migrated, we should be able to validate options at startup by using a startup filter that access the options. Given that, I think it'd be best if we do the migration first. Does that seem reasonable?

> There are still some options which are used before the validation happens

Hm this is a good point, I hadn't given much thought to this. Is `PathBase` in `Startup.Configure` the only option that's used before validation? Did you find others?

We can fix the `PathBase` option-before-validation issue by forking ASP.NET Core's `UsePathBase`. Let's do this in a separate pull request though!

> enums which are not valid are not handled.

Hm, do you think we should remove enums from options and use raw strings instead? We could add custom validations to ensure the string is one of the accepted values.

MarkZither:
I think the code has moved on from this so closing.

PR #459, opened 2020-02-05
add aliyun oss storage

chadzhao:
* Add BaGet.Aliyun project
* Add new StorageType : AliyunOSS


loic-sharma:
Hi thanks for opening this! It seems like it'd be possible to use the AWS S3 implementation to integrate with Aliyun's Object Storage Service. Do you know if there are any drawbacks to doing this? See: https://www.alibabacloud.com/help/doc-detail/64919.htm

src/BaGet.Aliyun/AliyunOSSService.cs
```diff
@@ -0,0 +1,90 @@
...
+        private string PrepareKey(string path)
+        {
+            return _prefix + path.Replace("\\", Separator);
+        }
+
+        public async Task<Stream> GetAsync(string path, CancellationToken cancellationToken = default)
+        {
+            try
+            {
+                return await Task.Run<Stream>(() =>
```
loic-sharma: Does the OSS SDK not support async? If so, let's make these methods non-async and return the results using `Task.FromResult(... return value here ...);` instead of using `Task.Run`

chadzhao: Thanks for let me know this. 

It does support async but only with IAsyncResult, I didn't have experience to due with IAsyncResult, to make BaGet run in docker (Aliyun Serverless Kubernates) ASAP, I made this in hurry.

I also tested to use AWS options for Aliyun OSS in docker, not work.

Now I make some improvement for async call.

src/BaGet.Aliyun/BaGet.Aliyun.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
loic-sharma: Let's add some metadata to make the package look a little better on nuget.org:

```xml
    <PackageTags>NuGet;Alibaba;Cloud</PackageTags>
    <Description>The libraries to host BaGet on Alibaba Cloud (Aliyun).</Description>
```

src/BaGet.Aliyun/BaGet.Aliyun.csproj
```diff
@@ -0,0 +1,15 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="aliyun.oss.sdk.core" Version="2.5.3" />
```
loic-sharma: Is this the right package? Alibaba's [documentation](https://www.alibabacloud.com/help/doc-detail/32086.htm?spm=a2c63.l28256.b99.487.646c5139tDvkNm#title-uh3-q4n-6lr) indicates that [Aliyun.OSS.SDK](https://www.nuget.org/packages/Aliyun.OSS.SDK/) or [Aliyun.OSS.SDK.NetCore ](https://www.nuget.org/packages/Aliyun.OSS.SDK.NetCore/) are the recommended packages. This package seems to be a third-party alternative that is no longer maintained.

src/BaGet.Aliyun/AliyunOSSService.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Aliyun.OSS;
+using BaGet.Aliyun.Configuration;
+using BaGet.Core;
+using Microsoft.Extensions.Options;
+
+namespace BaGet.Aliyun
+{
+    public class AliyunOSSService : IStorageService
```
loic-sharma: What do you think of the name `AliyunStorageService`? I'd rather avoid acronyms if possible

```suggestion
    public class AliyunOSSService : IStorageService
```

src/BaGet.Aliyun/Configuration/AliyunOSSOptions.cs
```diff
@@ -0,0 +1,22 @@
+using System.ComponentModel.DataAnnotations;
+using BaGet.Core;
+
+namespace BaGet.Aliyun.Configuration
+{
+    public class AliyunOSSOptions
```
loic-sharma: What do you think of the name `AliyunStorageOptions`?

```suggestion
    public class AliyunStorageOptions
```

src/BaGet.Aliyun/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+using Aliyun.OSS;
+using BaGet.Aliyun.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+
+namespace BaGet.Aliyun.Extensions
+{
+    public static class ServiceCollectionExtensions
+    {
+        public static IServiceCollection AddAliyunOssStorageService(this IServiceCollection services)
```
loic-sharma:
```suggestion
        public static IServiceCollection AddAliyunStorageService(this IServiceCollection services)
```

src/BaGet.Core/Configuration/StorageOptions.cs
```diff
@@ -12,5 +12,6 @@ public enum StorageType
         AwsS3 = 2,
         GoogleCloud = 3,
         Null = 4,
+        AliyunOSS = 5,
```
loic-sharma: For consistency let's lowercase the `ss` here:

```suggestion
        AliyunOss = 5,
```

src/BaGet.Aliyun/BaGet.Aliyun.csproj
chadzhao: Aliyun.OSS.SDK is for .net framework, Aliyun.OSS.SDK.NetCore is for .net core. But we need for .net standard. Aliyun go slow for .net because most of them use Java :-)

loic-sharma: Aliyun.OSS.SDK.NetCore seems to support .NET Standard: https://github.com/aliyun/aliyun-oss-csharp-sdk/blob/7efc8884e12bbc1accdc3fa1ee22d2bc3cec4717/sdk/aliyun-oss-sdk-dotnetcore.csproj#L12

chadzhao: Yes, I checked the source code and already switched to that sdk.

The name of the SDK misleads me. Seems like they need pay more attention on .net :-)

loic-sharma:
This is great work. Thank you @chadzhao!

PR #461, opened 2020-02-05
Allow the use of secret files for configuration

GotenXiao:
Allows users to use Docker secrets, or bind-mounted plain text files to provide settings values.

ASP.NET Documentation available [here](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-3.1#key-per-file-configuration-provider).

This also allows the prevention of exposing sensitive values in environment variables.

Both `AddKeyPerFile` calls appear to be necessary to make this work in all cases at present, pending the merge of `WebHostBuilder` and `HostBuilder` implementation.

An example `docker-compose.yml` file that makes use of this functionality:
```yaml
version: '2'

services:
  baget:
    build: .
    image: baget:secrets-configuration
    volumes:
      - ./data/secrets/connectionstring:/run/secrets/Database__ConnectionString:ro
      - ./data/secrets/apikey:/run/secrets/ApiKey:ro
      - ./data/db/:/db/
      - ./data/packages:/var/baget/packages
    environment:
      - Database__Type=Sqlite
      - Mirror__Enabled=true
      - Mirror__PackageSource=https://api.nuget.org/v3/index.json
      - Storage__Type=FileSystem
      - Storage__Path=/var/baget/packages
    ports:
      - '8000:80'
```


loic-sharma:
Thank you for opening this pull request and for providing a great description. I need to do a little bit of learning on this topic, but this looks great already!

src/BaGet/Program.cs
```diff
@@ -71,6 +71,8 @@ await provider
                     {
                         config.SetBasePath(root);
                     }
+
+                    config.AddKeyPerFile("/run/secrets", optional: true);
```
loic-sharma: What do you think of doing similar to the `BAGET_CONFIG_ROOT` environment variable - something like `BAGET_KEY_PER_FILE_PATH`. If the environment variable is set, we call `config.AddKeyPerFile(...);`. Does that seem reasonable?

GotenXiao: I can see the utility of allowing the directory to be configurable, but default to `/run/secrets`.

The `optional: true` parameter means the path is ignored if not present, so there should be no harm in calling this in all environments (no options values are provided in the event of path nonexistence or no files being present in the directory, and no exceptions are thrown by the provider). [Docker Secrets](https://docs.docker.com/engine/swarm/#how-docker-manages-secrets) and similar environments (including the now-legacy Rancher 1.6) use `/run/secrets` for secret value mountpoints by convention. Kubernetes environments are more flexible with the ability to select a mountpoint in a pod for a secret value.

One potential to note is that Kubernetes does support updating secret values in-situ on a deployed pod ([docs](https://kubernetes.io/docs/concepts/configuration/secret/#mounted-secrets-are-updated-automatically), so using `reload: true` as well might be useful - but that then requires the project to be aware of configuration fluidity, and use either `IOptionsMonitor<T>` for singleton services or `IOptionsSnapshot<T>` for scoped/transient services to receive the updated values.

GotenXiao:
Rebased on latest master.

Issue #458, opened 2020-02-05
Cannot delete packages b/c version is incorrectly case sensitive 


jcoburn:
### To Reproduce
1. Using this version of BaGet BaGet.dll reports version 1.0.0.0 2/5/2020, this is latest download 
2. Run this code 
C:\LocalCache> DotNet nuget push LandisGyr.AMI.ReadingTypes.1.0.0-Alpha0002.nupkg --api-key landisgyr --source http://localhost:5000/v3/index.json --symbol-source http://localhost:5000/v3/index.json --no-service-endpoint                 info : Pushing LandisGyr.AMI.ReadingTypes.1.0.0-Alpha0002.nupkg to 'http://localhost:5000/api/v2/package'...
info :   PUT http://localhost:5000/api/v2/package/
info :   Created http://localhost:5000/api/v2/package/ 876ms
info : Your package was pushed.
PS C:\LocalCache> DotNet nuget push LandisGyr.AMI.ReadingTypes.1.0.0-Alpha0003.nupkg --api-key landisgyr --source http://localhost:5000/v3/index.json --symbol-source http://localhost:5000/v3/index.json --no-service-endpoint                 info : Pushing LandisGyr.AMI.ReadingTypes.1.0.0-Alpha0003.nupkg to 'http://localhost:5000/api/v2/package'...
info :   PUT http://localhost:5000/api/v2/package/
info :   Created http://localhost:5000/api/v2/package/ 243ms
info : Your package was pushed.
PS C:\LocalCache> dotnet nuget delete LandisGyr.AMI.ReadingTypes 1.0.0-Alpha0002 --source http://localhost:5000/v3/index.json  --api-key landisgyr                                                                                              LandisGyr.AMI.ReadingTypes 1.0.0-Alpha0002 will be deleted from the 'http://localhost:5000/api/v2/package'. Would you like to continue? (y/N)
y
warn : Deleting LandisGyr.AMI.ReadingTypes 1.0.0-Alpha0002 from the 'http://localhost:5000/api/v2/package'.
info :   DELETE http://localhost:5000/api/v2/package/LandisGyr.AMI.ReadingTypes/1.0.0-Alpha0002
info :   NotFound http://localhost:5000/api/v2/package/LandisGyr.AMI.ReadingTypes/1.0.0-Alpha0002 238ms
error: Response status code does not indicate success: 404 (Not Found).
PS C:\LocalCache>        

Delete failed.

output from DotNet baget.dll
[...]

PR #466, opened 2020-02-08
Make the "Version" column case insensitive

loic-sharma:
## Background

BaGet is case sensitive on versions, which prevents unlisting/deleting a package that has uppercased letters in its version string (see https://github.com/loic-sharma/BaGet/issues/458). This only affects the SQLite and PostgreSQL drivers.

### SQLite

⚠SQLite does not support altering columns (see [this](https://docs.microsoft.com/en-us/ef/core/providers/sqlite/limitations#migrations-limitations)). Customers will need to recreate their database and then reupload all their packages to get this fix:

1. Delete the SQLite database: `rm baget.db`
1. Run BaGet to recreate the database
1. Reupload all the packages:
    1. Navigate to your storage folder: `cd packages`
    1. Push all packages: `dotnet nuget push .\**\*.nupkg --source http://localhost:5000/v3/index.json`

### Implementation

<details>
<summary>Test plan</summary>

* [x] PostgreSQL
    * [X] Test on a fresh database
    * [x] Test on an existing database after migrating
* [x] SQLite
    * [X] Test on a fresh database
    * [x] Test migration no-ops on an existing database
    * [x] Add instructions on how to rebuild SQLite database 
</details>


src/BaGet.Database.PostgreSql/BaGet.Database.PostgreSql.csproj
```diff
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="3.1.0" />
+    <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="3.1.1.2" />
```
loic-sharma: This package upgrade is necessary as npgsql's identity migration is broken on v3.1.0. This was fixed in 3.1.1.1.

src/BaGet.Database.Sqlite/Migrations/20180804082808_Initial.cs
```diff
@@ -32,7 +32,7 @@ protected override void Up(MigrationBuilder migrationBuilder)
                     RepositoryType = table.Column<string>(maxLength: 100, nullable: true),
                     Tags = table.Column<string>(maxLength: 4000, nullable: true),
                     RowVersion = table.Column<byte[]>(rowVersion: true, nullable: true),
-                    Version = table.Column<string>(maxLength: 64, nullable: false)
+                    Version = table.Column<string>(type: "TEXT COLLATE NOCASE", maxLength: 64, nullable: false)
```
loic-sharma: TODO: Add a comment about this hack

Issue #463, opened 2020-02-07
Error pushing huge packages on IIS (413 Payload Too Large)


Voxel7:
I'm unable to push huge packages on BaGet installed on IIS.
It seems related to #88.

I didn't figure out by trial and error the maximum allowable size. I'm able to reproduce the bug with packages of at least 41mb.

Push command:
```batch
C:\> nuget.exe push -source http://myserver:5000/v3/index.json MyHugePackage.nupkg
AVERTISSEMENT : No API Key was provided and no API Key could be found for 'http://myserver:5000/api/v2/package'. To save an API Key for a source use the 'setApiKey' command.
Pushing MyHugePackage.nupkg to 'http://myserver:5000/api/v2/package'...
  PUT http://myserver:5000/api/v2/package/
  RequestEntityTooLarge http://myserver:5000/api/v2/package/ 25948ms
Le code d'état de réponse n'indique pas la réussite : 413 (Payload Too Large).
```

Output from the log file:
```log
fail: Microsoft.AspNetCore.Server.IIS.Core.IISHttpServer[3]

      Unexpected exception in "IISHttpContext.ReadBody".

Microsoft.AspNetCore.Server.IIS.BadHttpRequestException: Request body too large.

   at Microsoft.AspNetCore.Server.IIS.BadHttpRequestException.Throw(RequestRejectionReason reason)

   at Microsoft.AspNetCore.Server.IIS.Core.IISHttpContext.ReadBody()

fail: Microsoft.AspNetCore.Server.IIS.Core.IISHttpServer[2]

      Connection ID "18086456109425492159", Request ID "800000c0-0001-fb00-b63f-84710c7967bb": An unhandled exception was thrown by the application.

Microsoft.AspNetCore.Server.IIS.BadHttpRequestException: Request body too large.

   at Microsoft.AspNetCore.Server.IIS.BadHttpRequestException.Throw(RequestRejectionReason reason)

   at Microsoft.AspNetCore.Server.IIS.Core.IISHttpContext.ReadBody()

   at System.IO.Pipelines.PipeCompletion.ThrowLatchedException()

   at System.IO.Pipelines.Pipe.GetReadResult(ReadResult& result)

   at System.IO.Pipelines.Pipe.GetReadAsyncResult()

   at Microsoft.AspNetCore.Server.IIS.Core.IISHttpContext.ReadAsync(Memory`1 memory, CancellationToken cancellationToken)

[...]

PR #473, opened 2020-02-13
Allow maximum published package size to be configured when using IIS

RichardSinden:
Now sets the MaxRequestBodySize for IIS on startup.

 * I've modified the startup code to set the IISServerOptions based on configuration in the appsettings.json file, with a default entry for MaxRequestBodySize with a maximum size of 250MB.
 * I've updated the documentation

Addresses https://github.com/loic-sharma/BaGet/issues/463

### DIFF
--- a/src/BaGet/Extensions/IServiceCollectionExtensions.cs
+++ b/src/BaGet/Extensions/IServiceCollectionExtensions.cs
@@ -20,6 +20,7 @@ using BaGet.Gcp.Configuration;
 using BaGet.Gcp.Extensions;
 using BaGet.Gcp.Services;
 using BaGet.Protocol;
+using Microsoft.AspNetCore.Builder;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
@@ -47,6 +48,7 @@ namespace BaGet.Extensions
             services.ConfigureAws(configuration);
             services.ConfigureGcp(configuration);
             services.ConfigureAliyunOSS(configuration);
+            services.ConfigureIis(configuration);
 
             if (httpServices)
             {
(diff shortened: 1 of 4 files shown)


src/BaGet/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -191,6 +193,15 @@ public static IServiceCollection AddBaGetContext(this IServiceCollection service
 
             return services;
         }
+        
+        public static IServiceCollection ConfigureIis(
+            this IServiceCollection services,
+            IConfiguration configuration)
+        {
+            services.ConfigureAndValidate<IISServerOptions>(configuration.GetSection(nameof(IISServerOptions)));
```
loic-sharma: I'd like to keep the appsettings.json file as light as possible. What do you think of setting the default of 250MB here? I think this should work:

```csharp
// IIS defaults upload sizes to ~30MB. Increase it to 250MB.
// See https://github.com/aspnet/Announcements/issues/267
services.Configure<IISServerOptions>(iis=>
{
    iis.MaxRequestBodySize = 262144000
});

services.ConfigureAndValidate<IISServerOptions>(configuration.GetSection(nameof(IISServerOptions)));
```

I believe this should default to 250MB, yet let customers override that value by adding a `IISServerOptions` option.

src/BaGet/appsettings.json
```diff
@@ -22,6 +22,10 @@
     "PackageSource": "https://api.nuget.org/v3/index.json"
   },
 
+  "IISServerOptions": {
+    "MaxRequestBodySize": 262144000
+  },
```
loic-sharma: If you go with my comment above https://github.com/loic-sharma/BaGet/pull/473/files#r378645236, please remove these lines since they are no longer necessary for the 250MB default.

loic-sharma:
Thank you for figuring this problem out. I'm a bit of an IIS noob so I really appreciate the help! :)

RichardSinden:
Yep, I agree, I wasn't entirely comfortable with always having config for IIS in the appsettings even when you are not using IIS. Your solution is better and I'll update it when I get the chance later, although I was going to wait and see if there was more information on this comment https://github.com/loic-sharma/BaGet/issues/463#issuecomment-585573492, in case we don't need to do anything at all.

softlion:
Is this MaxRequestBodySize not enough ?

```
  .ConfigureKestrel((context, options) =>
    {
        options.Limits.MaxRequestBodySize = ...;
    });
```

RichardSinden:
> 
> 
> Is this MaxRequestBodySize not enough ?
> 
No, this setting doesn't seem to have any effect when hosting in IIS. 

@loic-sharma, I've updated the PR as suggested to set a default in code and remove the appsettings.json default section. I've updated the documentation to make reference to the IISServerOptions section in the Microsoft docs.

loic-sharma:

PR #491, opened 2020-03-18
Fix #490 - package last updated date

### LINKED ISSUE #490: UI: Package last updated date - shown in US format in other locales
When viewing a package, the `Last updated` field shows the date in US format (`MM/DD/YYYY`), rather than the locale's format. This format [appears to be hard-coded](https://github.com/loic-sharma/BaGet/blob/master/src/BaGet.UI/src/DisplayPackage/Versions.tsx#L101).

### Describe the solution you'd like

Ideally the date should be rendered using the browser locale settings. This can be done using [`Date.prototype.toLocaleDateString`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString)

euantorano:
Fixes #490 by using `Date.prototype.toLocaleDateString()` to format date values used to show when a package was last updated.

Addresses https://github.com/loic-sharma/BaGet/issues/490

### DIFF
--- a/src/BaGet.UI/src/DisplayPackage/Versions.tsx
+++ b/src/BaGet.UI/src/DisplayPackage/Versions.tsx
@@ -99,6 +99,6 @@ export class Versions extends React.Component<IVersionsProps, IVersionsState> {
   private showFewerVersions = () => this.setState({ showAll: false });
 
   private dateToString(date: Date): string {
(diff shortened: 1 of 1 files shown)


loic-sharma:
Awesome! Thanks for fixing this :)

PR #495, opened 2020-03-25
Fix request pipeline for custom PathBase settings

daniel-lerch:
Summary of the changes

 * `UsePathBase` only works if applied before any file servers
 * This is just a quick fix before #388 is merged


loic-sharma:
Thanks for the contribution!

PR #501, opened 2020-04-05
Let environment variables overwrite configuration file settings

jeroenlandheer:
In case of ambiguity, environment variables are overwritten by configuration file settings, this should be the other way around. (Environment variables should have a higher priority.)

This PR changes the order of applying configuration variables.


loic-sharma:
Thanks for the contribution!

PR #511, opened 2020-05-07
PathBase support for React UI

### LINKED ISSUE #504: Frontend doesn't use PathBase
### Describe the bug

I have a reverse proxy (nginx) in front of baget that forwards requests to http://host/baget to the docker container. The docker container is configured with `-e PathBase=/baget`. In VS, download works fine, but the browser UI at http://host/baget fails to load correctly:

Static resources are not loaded correctly, but that can be fixed by forwarding requests with the referer http://host/baget to the docker container. However, links in the UI (e.g. Packages) at the top link to http://host, i.e. without the correct path. The API call that tries to fetch the list of packages also goes to http://host instead of http://host/baget.

### To Reproduce
1. Set up reverse proxy, e.g. with nginx
```
server {
        listen 80;
        location /nuget/ {
            proxy_pass http://host.docker.internal:8080/; # assuming baget is reachable on this URI
        }
}
```
2. Browse `http://localhost/nuget/` in a browser

daniel-lerch:
Summary of the changes

 * Use placeholders for PathBase in React source code
 * Implement a custom `ISpaStaticFileProvider` to replace these placeholders

Addresses #276
Fixes #504

This is an improved version of #388 by @szwenni

 * Better code style
 * Efficient caching with lazy reload
 * Registered in `ConfigureHttpServices`


src/BaGet.UI/src/config.tsx
```diff
@@ -1,13 +1,22 @@
 let config = {
-  apiUrl: "__BAGET_PLACEHOLDER_API_URL__"
+  apiUrl: "__BAGET_PLACEHOLDER_API_URL__",
+  baseUrl: "__BAGET_PATH_BASE_PLACEHOLDER__"
```
loic-sharma, line 3: This should start with `__BAGET__PLACEHOLDER_`, otherwise line 10 doesn't work.

```suggestion
  baseUrl: "__BAGET_PLACEHOLDER_PATH_BASE__"
```

src/BaGet.Core/Configuration/BaGetOptions.cs
```diff
@@ -16,6 +16,10 @@ public class BaGetOptions
         /// </summary>
         public string PathBase { get; set; }
 
+        /// <summary>
+        /// The application Host without port number. If not set all hosts will be allowed. See https://github.com/aspnet/Announcements/issues/295
+        /// </summary>
+        public string HostUrl { get; set; }
```
loic-sharma, line 22: Let's align this with the `ForwardedHeadersOptions.AllowedHosts` property by renaming this to `AllowedHosts` and making it a list of strings.

loic-sharma:
This looks great! Thank you for starting this. I left some comments but I need to dig deeper as I don't fully understand the static file providers just yet.

src/BaGet.Hosting/FileProvider/BaGetSpaStaticFileProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+            var rootPath = "BaGet.UI/build";
+            if (environment.IsDevelopment())
+            {
+                rootPath = "../" + rootPath;
+            }
+
+            var combinedRootPath = new DirectoryInfo(Path.Combine(environment.ContentRootPath, rootPath));
+            if (combinedRootPath.Exists)
+            {
+                var pathBase = options.Value.PathBase;
```
loic-sharma, line 26: You can modify BaGet's configs as the app is running and it will automatically refresh them. This is capturing the `PathBase` at startup. Can we pass the options down to the file provider and let that read the latest `PathBase` each time it serves a file?

src/BaGet.Hosting/FileProvider/BaGetUIFileProvider.cs
```diff
@@ -0,0 +1,136 @@
...
+            _matcher = new Matcher().AddInclude("**/index.html").AddInclude("**/*.js");
+            _fileCache = new ConcurrentDictionary<string, CachedFileInfo>();
+        }
+
+        public IDirectoryContents GetDirectoryContents(string subpath)
+        {
+            return new DirectoryContents(this, new DirectoryInfo(Path.Combine(_root.FullName, subpath.TrimStart('/'))));
+        }
+
+        public IFileInfo GetFileInfo(string subpath)
```
loic-sharma, line 36: I'd prefer it if this feature was a little more "pay for play". Right now we always cache files in memory even the `PathBase` configuration wasn't set. It'd be better if we don't cache or replace `__BAGET_PLACEHOLDER_...` if `PathBase` isn't configured.

src/BaGet.Hosting/FileProvider/BaGetUIFileProvider.cs
```diff
@@ -0,0 +1,136 @@
...
+        private readonly PhysicalFilesWatcher _watcher;
+        private readonly Matcher _matcher;
+        private readonly ConcurrentDictionary<string, CachedFileInfo> _fileCache;
+
+        public BaGetUIFileProvider(DirectoryInfo root, string pathBase)
+        {
+            _root = root;
+            _pathBase = pathBase;
+            _watcher = new PhysicalFilesWatcher(root.FullName, new FileSystemWatcher(root.FullName), true);
+            _matcher = new Matcher().AddInclude("**/index.html").AddInclude("**/*.js");
```
loic-sharma, line 27: Do we need to match the `index.html` file? I believe only `.js` files have placeholders today. We might be able to replace this with a `Path.GetExtensions(subpath) == ".js"`

loic-sharma:
I think we can simplify `BaGetUIFileProvider` by using the default `ISpaStaticFileProvider`'s `IFileProvider`. I made a prototype of this here: https://github.com/loic-sharma/BaGet/compare/loshar-path-base

src/BaGet.Hosting/FileProvider/BaGetUIFileProvider.cs
daniel-lerch: I know three options to implement a custom pathbase with ASP.NET and Node:
1. Replace `__BAGET_PLACEHOLDER_...` in every affected file and cache them. I would not expect any performance penalty as files are kept in memory. An instance of BaGet occupies ~100 MB RAM and caching all JavaScript files would another ~700 KB.
2. Replace `__BAGET_PLACEHOLDER_...` only in `index.html` and implement code in `config.tsx` to dynamically replace `__BAGET_PLACEHOLDER_...` with an empty string. This would reduce the additional memory usage for standard deployments to less than 3 KB.
3. Omit placeholders in `index.html` and use the concept from 2. for JavaScript. If a pathbase is configured we could edit `index.html` with regular expressions. This would require a full test coverage because we would have to rely on implementation details of the Node compiler.

If you know a better way to implement a custom pathbase please describe it to me or tell me which of these three options would like to have in BaGet.

src/BaGet.Core/Configuration/BaGetOptions.cs
daniel-lerch: The way how BaGet configures its forwarded headers middleware has changed. I want to keep this pull request simple so if someone needs this feature, please open an issue.

src/BaGet.Hosting/FileProvider/BaGetUIFileProvider.cs
daniel-lerch: @loic-sharma What's your opinion on that?

src/BaGet.Hosting/FileProvider/BaGetUIFileProvider.cs
daniel-lerch: You're absolutely right.

daniel-lerch:
> I think we can simplify `BaGetUIFileProvider` by using the default `ISpaStaticFileProvider`'s `IFileProvider`. I made a prototype of this here: https://github.com/loic-sharma/BaGet/compare/loshar-path-base

Yes, we can replace `BaGetSpaStaticFileProvider` with the default `ISpaStaticFileProvider`. I like the way you registered it in your prototype and will adapt that soon.

However, using the default `IFileProvider` which is actually a `PhysicalFileProvider` as a backend for our own file provider leads to a problem:
`IFileProvider` exposes a function `GetDirectoryContents` which contains an `IEnumerable<IFileInfo>` which are physical files and not the replaced and cached ones. It might be that ASP.NET Core only uses this method for directory indices which are not necessary for UI files but I don't like to rely on such implementation details.

src/BaGet/Startup.cs
```diff
@@ -113,10 +106,14 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
                 spa.Options.SourcePath = "../BaGet.UI";
 
-                if (env.IsDevelopment())
-                {
-                    spa.UseReactDevelopmentServer(npmScript: "start");
-                }
+                //if (env.IsDevelopment())
+                //{
+                //    spa.UseReactDevelopmentServer(npmScript: "start");
+                //}
```
daniel-lerch, line 112: This React development server is really a headache for me. If it is enabled in all development environments, integration tests for my SPA file provider with custom path base fail. In addition to that we have a second middleware for serving SPA static files which should be disabled when using the React server.

My idea for a workaround would be to add a setting  `BaGetOptions.DisableReactDevelopmentServer`. Per default, all development environments would use the React server and unit tests would disabled it with an in-memory configuration. Any other ideas?

daniel-lerch:
This repository has been inactive for too long to keep this PR up to date.

There are better ways to implement a custom base path for Vue.js which I found out working on [another project](https://github.com/daniel-lerch/korga). I am sure such an approach would work for BaGet and React, too.

Issue #303, opened 2019-08-12
Add a landing page when there are no packages


loic-sharma:
### Is your feature request related to a problem? Please describe.
BaGet shows an empty home page when you start it for the first time.

### Describe the solution you'd like
There should be a landing page that shows you how to get started. This page should have links to the upload page and the documentation.

codenamejason:
Is this actually still open?

loic-sharma:
Yup. Here are the two landing pages we could improve:

### Home page, if no packages have been uploaded

When you run BaGet for the first time, you are greeted with:

![image](https://user-images.githubusercontent.com/737941/74048796-be47a380-4987-11ea-8723-899f45399471.png)

This page feels cold and empty, we should make it fun and inviting! Also, I would like to provide a link to the upload page, a link to the documentation website, and a link to BaGet's GitHub issue tracker.

### Package page, if the package doesn't exist

If you search for a package that doesn't exist, BaGet gives a rather unhelpful message:

![image](https://user-images.githubusercontent.com/737941/74048561-4e391d80-4987-11ea-98e1-ef0ee74ba14d.png)

This error message is correct but not particularly helpful! We should add a link to the package on nuget.org, provide a link to the documentation, and suggest asking for help using BaGet's GitHub issue tracker.

sipmann:
Is this issue under development? Or can I take it :) ?

loic-sharma:
No one's working on this right now, feel free to take it! Please let me know if you have any questions :)

sipmann:
Cool, I'm already working on it. It's something like the prints bellow that you have in mind? 

* Documentation links to https://loic-sharma.github.io/BaGet
* Issues/GitHub Project links to https://github.com/loic-sharma/BaGet/issues

## Index/Search without packages on the list
![image](https://user-images.githubusercontent.com/545754/84093611-b8f77580-a9c8-11ea-815c-b687a2ad0f84.png)

## Package page
![image](https://user-images.githubusercontent.com/545754/84093604-b268fe00-a9c8-11ea-8197-f940a2f3d532.png)

loic-sharma:
Yup that looks great!

sipmann:
The Pull request didn't closed the issue. Next time I'll use the proper keyword.

loic-sharma:
No worries. Thanks for the good work @sipmann!

PR #522, opened 2020-06-09
Improvement of empty lists and 404 pages

sipmann:

 * Added a message on the index/search page when there is no package on the list.
 * Added a message on the package page where the package can't be found.

Addresses https://github.com/loic-sharma/BaGet/issues/303

### DIFF
--- a/src/BaGet.UI/src/DisplayPackage/DisplayPackage.tsx
+++ b/src/BaGet.UI/src/DisplayPackage/DisplayPackage.tsx
@@ -208,8 +208,8 @@ class DisplayPackage extends React.Component<IDisplayPackageProps, IDisplayPacka
         <div>
           <h2>Oops, package not found...</h2>
           <p>Could not find package '{this.id}'.</p>
-          <p>You can take a look at the <a href={`https://www.nuget.org/packages/${this.id}`} target="_blank">nuget.org</a> package.</p>
-          <p>If you think there's a problem? Consider taking a look at our <a href="https://loic-sharma.github.io/BaGet/" target="_blank">Documentation</a> or ask for help at our <a href="https://github.com/loic-sharma/BaGet/issues" target="_blank">GitHub Project</a></p>
+          <p>You can try searching on <a href={`https://www.nuget.org/packages?q=${this.id}`} target="_blank" rel="noopener noreferrer">nuget.org</a> package.</p>
+          <p>Think there's a problem? Consider taking a look at our <a href="https://loic-sharma.github.io/BaGet/" target="_blank" rel="noopener noreferrer">documentation</a> or asking for help on our <a href="https://github.com/loic-sharma/BaGet/issues" target="_blank" rel="noopener noreferrer">GitHub project</a></p>
         </div>
       );
     } else {
--- a/src/BaGet.UI/src/DisplayPackage/DisplayPackage.tsx
+++ b/src/BaGet.UI/src/DisplayPackage/DisplayPackage.tsx
@@ -205,7 +205,12 @@ class DisplayPackage extends React.Component<IDisplayPackageProps, IDisplayPacka
       );
     } else if (!this.state.package) {
       return (
-        <div>Could not find package '{this.id}'.</div>
+        <div>
+          <h2>Oops, package not found...</h2>
+          <p>Could not find package '{this.id}'.</p>
+          <p>You can take a look at the <a href={`https://www.nuget.org/packages/${this.id}`} target="_blank">nuget.org</a> package.</p>
+          <p>If you think there's a problem? Consider taking a look at our <a href="https://loic-sharma.github.io/BaGet/" target="_blank">Documentation</a> or ask for help at our <a href="https://github.com/loic-sharma/BaGet/issues" target="_blank">GitHub Project</a></p>
+        </div>
       );
     } else {
       return (
--- a/src/BaGet.UI/src/SearchResults.tsx
+++ b/src/BaGet.UI/src/SearchResults.tsx
@@ -156,8 +156,67 @@ class SearchResults extends React.Component<ISearchResultsProps, ISearchResultsS
             />
           </div>
         </form>
-
-        {this._renderSearch()}
+        
+        {(() => {
+          if (this.state.items.length > 0) {
+            return this.state.items.map(value => (
+              <div key={value.id} className="row search-result">
+                <div className="col-sm-1 hidden-xs hidden-sm">
+                  <img
+                    src={value.iconUrl || DefaultPackageIcon}
+                    className="package-icon img-responsive"
+                    onError={this.loadDefaultIcon}
+                    alt="The package icon" />
+                </div>
+                <div className="col-sm-11">
+                  <div>
+                    <Link to={`/packages/${value.id}`} className="package-title">{value.id}</Link>
+                    <span>by: {value.authors}</span>
+                  </div>
+                  <ul className="info">
+                    <li>
+                      <span>
+                        <Icon iconName="Download" className="ms-Icon" />
+                        {value.totalDownloads.toLocaleString()} total downloads
+                      </span>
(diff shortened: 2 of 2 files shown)


src/BaGet.UI/src/SearchResults.tsx
```diff
@@ -156,48 +156,9 @@ class SearchResults extends React.Component<ISearchResultsProps, ISearchResultsS
...
-                }
-              </ul>
-              <div>
-                {value.description}
-              </div>
-            </div>
-          </div>
-        ))}
+
+        {this._renderSearch()}
```
loic-sharma, line 160: I'd rather avoid splitting up the HTML here if possible. Is it possible to do something like:

```tsx
{(() => {
  if (this.state.items.length > 0) {
    // Show results
  } else {
    // Show empty message
  }
})()}
```

src/BaGet.UI/src/SearchResults.tsx
```diff
@@ -298,6 +259,68 @@ class SearchResults extends React.Component<ISearchResultsProps, ISearchResultsS
       this.state.packageType,
       this.state.targetFramework);
   }
+
+  private _renderSearch() {
+    console.log(this.state.items);
```
loic-sharma, line 264: Please remove this line

src/BaGet.UI/src/SearchResults.tsx
```diff
@@ -298,6 +259,68 @@ class SearchResults extends React.Component<ISearchResultsProps, ISearchResultsS
...
+    {
+      return (
+        <div>
+          <h2>Oops, nothing here...</h2>
+          <p>
+            It looks like there's no package here to see. Take a look below for useful links.
+          </p>
+          <p><Link to="/upload">Upload a package</Link></p>
+          <p><a href="https://loic-sharma.github.io/BaGet/" target="_blank">BaGet Documentation</a></p>
+          <p><a href="https://github.com/loic-sharma/BaGet/issues" target="_blank">BaGet Issues</a></p>
```
loic-sharma, line 319: Casing suggestion:

```suggestion
          <p><a href="https://loic-sharma.github.io/BaGet/" target="_blank">BaGet documentation</a></p>
          <p><a href="https://github.com/loic-sharma/BaGet/issues" target="_blank">BaGet issues</a></p>
```

src/BaGet.UI/src/DisplayPackage/DisplayPackage.tsx
```diff
@@ -205,7 +205,12 @@ class DisplayPackage extends React.Component<IDisplayPackageProps, IDisplayPacka
       );
     } else if (!this.state.package) {
       return (
-        <div>Could not find package '{this.id}'.</div>
+        <div>
+          <h2>Oops, package not found...</h2>
+          <p>Could not find package '{this.id}'.</p>
+          <p>You can take a look at the <a href={`https://www.nuget.org/packages/${this.id}`} target="_blank">nuget.org</a> package.</p>
+          <p>If you think there's a problem? Consider taking a look at our <a href="https://loic-sharma.github.io/BaGet/" target="_blank">Documentation</a> or ask for help at our <a href="https://github.com/loic-sharma/BaGet/issues" target="_blank">GitHub Project</a></p>
```
loic-sharma, line 212:
```suggestion
          <p>You can try searching on <a href={`https://www.nuget.org/packages?q=${this.id}`} target="_blank">nuget.org</a>.</p>
          <p>Think there's a problem? Consider taking a look at our <a href="https://loic-sharma.github.io/BaGet/" target="_blank"dDocumentation</a> or asking for help on our <a href="https://github.com/loic-sharma/BaGet/issues" target="_blank">GitHub project</a></p>
```

loic-sharma:
Thanks for doing this, it looks great already! I left some minor comments, I'll merge this in once you address them :)

src/BaGet.UI/src/SearchResults.tsx
sipmann: Yeah for sure. You mean take the code from `_renderSearch` back to the render and do the if inside it. Right?
```suggestion
      render() {
         ...
         <div className="form-group">
            <Checkbox
              defaultChecked={this.state.includePrerelease}
              onChange={this.onChangePrerelease}
              label="Include prerelease:"
              boxSide="end"
            />
          </div>
        </form>
        
        // {this._renderSearch()}
        if (this.state.items.length > 0) {
            this.state.items.map( i => ...
         } else { ...
```

sipmann: To be easier to understand. Something like this? https://gist.github.com/sipmann/9e266ca45333238ccfd9c14f033e2fa2

loic-sharma: Yup, that looks perfect

sipmann:
Awesome, I've just commited every suggestion :). ty for your patience

src/BaGet.UI/src/DisplayPackage/DisplayPackage.tsx
```diff
@@ -205,7 +205,12 @@ class DisplayPackage extends React.Component<IDisplayPackageProps, IDisplayPacka
       );
     } else if (!this.state.package) {
       return (
-        <div>Could not find package '{this.id}'.</div>
+        <div>
+          <h2>Oops, package not found...</h2>
+          <p>Could not find package '{this.id}'.</p>
+          <p>You can try searching on <a href={`https://www.nuget.org/packages?q=${this.id}`} target="_blank" rel="noopener noreferrer">nuget.org</a> package.</p>
```
loic-sharma, line 211: Nice job on these `rel` attributes, I didn't know these existed 👍

src/BaGet.UI/src/SearchResults.tsx
```diff
@@ -156,48 +156,68 @@ class SearchResults extends React.Component<ISearchResultsProps, ISearchResultsS
...
-                onError={this.loadDefaultIcon}
-                alt="The package icon" />
-            </div>
-            <div className="col-sm-11">
-              <div>
-                <Link to={`/packages/${value.id}`} className="package-title">{value.id}</Link>
-                <span>by: {value.authors}</span>
+        
+        {(() => {
+          if (this.state.items.length > 0) {
```
loic-sharma, line 161: Hm I noticed there's a small flaw with this logic. When we change the search input, we set `items` to an empty array, thereby causing that "Oops, nothing here..." message to flash while we load the new items. I think we should introduce a new `loading` property on the state object and change this line to:

```suggestion
          if (loading || this.state.items.length > 0) {
```

When we start loading new items using [`_loadItems`](https://github.com/loic-sharma/BaGet/blob/2ed9457a0c80e4755a224911f2509f44e8e5a8bf/src/BaGet.UI/src/SearchResults.tsx#L225), we should set `loading` to `true`. Once we complete loading items, we should set `loading` to `false`.

loic-sharma:
Ok I left one more comment to polish up the UI as we load search elements. I can't wait to get this merged in! :)

src/BaGet.UI/src/SearchResults.tsx
sipmann: Yeap, totally agreed

sipmann:
Almost done with the last suggestion. But I've found a bug at the way the search works. Fill a bug a solve in another pull request? I solve on this one?
It should be just a one line change

loic-sharma:
What's the bug? If it's just a one liner then we can probably add it to this change as well

sipmann:
At line 71, we are sending the previous input value, instead of the new one. So when we type the next letter, the previous one is sent.

```javascript
this._loadItems(
      prevProps.input,
```

search_bug

loic-sharma:
Oh nice catch! Feel free to fix it in this PR or in another one, both approaches are fine by me :)

loic-sharma:
Thanks for the awesome contribution :)

PR #532, opened 2020-06-11
Update frontend dependencies

SeppPenner:

 * Updated npm packages.

This error occured after updating: 
grafik

I'm not an expert for React, so I have no idea :D

### DIFF
--- a/src/BaGet.UI/package.json
+++ b/src/BaGet.UI/package.json
@@ -11,7 +11,7 @@
   "repository": "github:loic-sharma/BaGet",
   "private": true,
   "dependencies": {
-    "bootstrap": "3.3.7",
+    "bootstrap": "^3.4.1",
     "copy-text-to-clipboard": "^2.2.0",
     "office-ui-fabric-react": "^7.119.1",
     "react": "^16.13.1",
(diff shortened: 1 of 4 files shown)


loic-sharma:
Hm that error is weird. You should be able to get around it by changing this line:

https://github.com/loic-sharma/BaGet/blob/63d7633c070bcae5fd4faea3415568b170afe70e/src/BaGet.UI/src/Upload.tsx#L58

To:

```tsx
{'>' + value}
```

src/BaGet.UI/package.json
```diff
@@ -13,11 +13,11 @@
   "dependencies": {
     "bootstrap": "3.3.7",
```
loic-sharma, line 14: Could you bump `bootstrap` to 3.4.1 to address this: https://github.com/advisories/GHSA-9v3m-8fp8-mj99

This should do the trick: ` npm install bootstrap@3.4.1`

SeppPenner: Ok.

SeppPenner:
There are some more outdated packages `npm --list outdated`:

grafik

Should I update any of them as well?

SeppPenner:
This makes https://github.com/loic-sharma/BaGet/pull/274 not needed any longer, by the way.

loic-sharma:

PR #535, opened 2020-06-17
Ensures that the registration index is sorted by version

### LINKED ISSUE #534: Version order is not enforced
After migrating from _NuGet.Server_ to _BaGet_ ussing the [suggested script](https://loic-sharma.github.io/BaGet/tools/migrate/#migrate-from-nugetserver), we ended up with the _Versions_ section of each package in a more or less random order.

Digging up the code, I've found that the Versions views assumes the list is ordered following SemVer, and reverses it:

```tsx
  public render() {
    // Show the versions from greatest to smallest.
    // We clone the versions array to avoid modifying the props.
    // This assumes the versions are ordered by semver ascendingly.
    let versionsToRender = [...this.props.versions];
    versionsToRender.reverse();
    if (!this.state.showAll) {
      versionsToRender = versionsToRender.slice(0, Versions.defaultVisible);
    }
```

But the method `FindAsync` in `PackageService` does not enforce an order:

```csharp
        public async Task<IReadOnlyList<Package>> FindAsync(string id, bool includeUnlisted, CancellationToken cancellationToken)
        {
            var query = _context.Packages
                .Include(p => p.Dependencies)
                .Include(p => p.PackageTypes)
                .Include(p => p.TargetFrameworks)
                .Where(p => p.Id == id);

            if (!includeUnlisted)
            {
                query = query.Where(p => p.Listed);
            }

            return (await query.ToListAsync(cancellationToken)).AsReadOnly();
        }
```

So the packages end up ordered (indirectly) by `Key`. This happens to be the "correct" order, indirectly, if packages are published naturally during their lifetime, assuming each new version has a higher _SemVer_ than the previous.

But I think it would be better to enforce the assumed ordering in the service to avoid any ambiguity (like for example the one we got into after a migration). Not forcing an order could also break normal use cases: if someone publish a bugfix package for a previous (lower) major version, that version would be the first one shown in the _Versions_ section, breaking the ordering.

chaghi:
The registration index is ordered by version as expected by the UI.

Fixes #534

### DIFF
--- a/src/BaGet.Core/Metadata/RegistrationBuilder.cs
+++ b/src/BaGet.Core/Metadata/RegistrationBuilder.cs
@@ -16,7 +16,7 @@ namespace BaGet.Core
 
         public virtual BaGetRegistrationIndexResponse BuildIndex(PackageRegistration registration)
         {
-            var versions = registration.Packages.Select(p => p.Version).ToList();
+            var sortedPackages = registration.Packages.OrderBy(p => p.Version).ToList();
 
             // TODO: Paging of registration items.
             // "Un-paged" example: https://api.nuget.org/v3/registration3/newtonsoft.json/index.json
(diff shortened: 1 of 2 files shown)


loic-sharma:

Issue #520, opened 2020-06-02
Add version information to the Web UI


SeppPenner:
### Is your feature request related to a problem? Please describe.
The web UI lacks the version information of the BaGet server.

### Describe the solution you'd like
The web UI should display the version information of the BaGet server.

### Describe alternatives you've considered
None.

### Additional context
None.

loic-sharma:
This would definitely be useful. What do you think of adding a footer, like nuget.org's?

![image](https://user-images.githubusercontent.com/737941/84224940-b4978f00-aa92-11ea-8fce-0adb6f5cba97.png)

BaGet's version number would be useful on the footer. Maybe we could also link to documentation and GitHub?

SeppPenner:
Yes, that's what I was thinking of 👍

PR #537, opened 2020-06-18
Added basic footer to the page.

SeppPenner:

 * Added basic footer.

Fixes https://github.com/loic-sharma/BaGet/issues/520 once it's alright.


SeppPenner:
@loic-sharma Currently, it looks like this:

BaGet_Footer

So, the todos are obvious:

* What do we want to show there?
* How do we get the version information there (Best would be automatically from Git somehow)?
* How can I fix the logo sizing?
* Another logo is needed, we can't take the official nuget logo 😄 
* The footer needs to be fixed to the bottom (even if there are no packages yet)

Since I'm not really a JavaScript / React or CSS expert, can you give me some hints here on these issues, please? Thank you already 😅

src/BaGet.UI/src/App.tsx
```diff
@@ -23,18 +25,49 @@ class App extends React.Component<RouteComponentProps, IAppState> {
...
 
         {this._renderContent()}
       </div>
-    );
+      ),
+      (
+        <footer className="footer">
+          {this._renderFooter()}
+        </footer>
+      )];
```
loic-sharma, line 39: Let's move the  `footer` element to inside the `_renderFooter` method. Also, you should be able to move the `this._renderFooter() to be next to the `_renderContent`. Something like:

```jsx
return (
  <div>
    {this._renderNavigationBar()}

    {this._renderContent()}

    {this._renderFooter()}
  </div>
);
```

src/BaGet.UI/src/App.css
```diff
@@ -77,3 +77,70 @@
...
+  position: relative;
+  min-height: 1px;
+  padding-right: 15px;
+  padding-left: 15px;
+}
+
+.row {
+  margin-right: -15px;
+  margin-left: -15px;
+}
```
loic-sharma, line 115: I believe all of these are in the Bootstrap stylings. You shouldn't need to redefine them.

loic-sharma:
Thanks for sharing a screenshot!

> How do we get the version information there (Best would be automatically from Git somehow)?

You should be able to use this: https://stackoverflow.com/a/52079972/1123509

You will need to update the Azure DevOps build script (found [here](https://github.com/loic-sharma/BaGet/blob/master/.azure/pipelines/ci-official.yml)) to set this environment variable.

> * How can I fix the logo sizing?
> * Another logo is needed, we can't take the official nuget logo 😄

BaGet doesn't have a logo 😢, I would just skip the logo for now.

> * The footer needs to be fixed to the bottom (even if there are no packages yet)

TODO...

SeppPenner:
I will try this and see how it looks after the changes :)

src/BaGet.UI/src/App.tsx
SeppPenner: Yep, that's a good idea.

src/BaGet.UI/src/App.css
SeppPenner: I have tested that, it's true.

SeppPenner:
Todos:

* What do we want to show there?
* Get version information like described in the links
* Update build script
* The footer needs to be fixed to the bottom (even if there are no packages yet)

PR #539, opened 2020-06-22
Update the bindings in docker quickstart

otsenkun:
Update the bindings in docker quickstart

 * When running in a docker container and without changing any aspnetcore binding, the default port is `5000` and not `80` as mentionned in the documentation.


loic-sharma:
Thanks for your contribution! I'll redeploy the documentation website soon with your fix :)

Issue #548, opened 2020-06-29
Env var support missing?


ffandreassoroko:
### Describe the bug
We are using the Baget docker image and noticed in the last updates, that the env var support was dropped? 

### Additional context
also its not possible to change the url/port for the container with the env `ASPNETCORE_URLS`
(before it was 80 and now it is 5000)
i think if you would support this, you could also remove your "custom" url setting 
https://github.com/loic-sharma/BaGet/blob/master/src/BaGet/Program.cs#L85

downgraded back to : 636c38146791bf7bdaa297587e196508f732cd08

`.AddEnvironmentVariables()` would be great 👍 
https://github.com/loic-sharma/BaGet/blob/master/src/BaGet/Program.cs#L82

loic-sharma:
Thanks for reporting this! Also, thanks for the detailed bug report, it helps a lot :)

I think the easiest fix may be to revert https://github.com/loic-sharma/BaGet/pull/525 for now. I think I need to think on this a little more as it's not clear how to be solve this problem right now.

> ... before [the port] was 80 and now it is 5000

Ah this change was not intentional. Thanks for the heads up.

> also its not possible to change the url/port for the container with the env `ASPNETCORE_URLS`

This is happening because the `appsettings.json` file's `Url` config is overriding the URL provided by the `ASPNETCORE_URLS` environment variable.

loic-sharma:
I'm working on a fix here: https://github.com/loic-sharma/BaGet/pull/549. I'll let others review this in case they have concerns, but will merge this in sometime tomorrow.

loic-sharma:
Ok this should be fixed and a new Docker image should be created soon. Please let me know if you run into any additional issues when using the latest version of BaGet.

Please note that I recently did a massive refactoring of BaGet (https://github.com/loic-sharma/BaGet/pull/538); I tested this as thoroughly as possible but sweeping changes like this may have bugs.

ffandreassoroko:
before we update and getting again some problems... just a little question. 

is it still possible to configure baget over env ? ( `Storage__Type`, `Database__Type` etc.. )

loic-sharma:
Yup that should still work.

PR #549, opened 2020-06-30
Fix URL configuration issues

loic-sharma:
Fixes https://github.com/loic-sharma/BaGet/issues/548

Sadly the `urls` config is no longer as clean as @SeppPenner's change. However, this solution works with Docker scenarios. In the future I'll consider how to bring back the `Urls` top-level config without breaking Docker.

I also added the `--urls` option to mirror ASP.NET Core. For more information, see: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel?view=aspnetcore-3.1#endpoint-configuration

/cc @ffandreassoroko
/cc @SeppPenner

### DIFF
--- a/src/BaGet/Program.cs
+++ b/src/BaGet/Program.cs
@@ -76,19 +76,6 @@ namespace BaGet
                         options.Limits.MaxRequestBodySize = null;
                     });
 
-                    var config = new ConfigurationBuilder()
-                        .SetBasePath(Directory.GetCurrentDirectory())
-                        .AddJsonFile("appsettings.json", optional: true)
-                        .AddCommandLine(args)
-                        .Build();
-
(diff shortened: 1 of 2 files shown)


SeppPenner:
This issue was definitely introduced by the change I did, yes...

loic-sharma:
That’s alright, this is a very subtle bug that only manifests itself if you’re on Docker and use `ASPNETCORE_URLS`.

SeppPenner:
> That’s alright, this is a very subtle bug that only manifests itself if you’re on Docker and use ASPNETCORE_URLS.

I have only worked with docker a few times 😄 And honestly haven't tried to run BaGet in Docker at all. But sure, there will be a workaround for this somehow :)

PR #560, opened 2020-07-21
allow to filter search result by regular expression

DavidHamburg:
Hi,

thanks for your great nuget alternative.

We have several instances of baget for different purposes:

  a) internal packages
  b) caching packages from nuget.org

We would like to prevent that the cache server gives use packages with a certain name schema to avoid clashes with internal packages. Lets say we use *xyz* as prefix for packages e.g. xyz.xmlparser and someone uploads another package with the same name to nuget.org, we must be sure we get always our package.
Another use-case would be to block malicious packages.

The current behavior does not change unless the regular expression has been configured.

Summary:
 * Add a new option to blacklist packages by regular expression

--
Best regards,

David


loic-sharma:
Hi, thanks for opening this pull request! Why do you want to filter packages at the search level? Wouldn't it be better to filter results in the [mirror service](https://github.com/loic-sharma/BaGet/blob/master/src/BaGet.Core/Mirror/MirrorService.cs), which is what discovers packages from an upstream NuGet server like nuget.org?

docs/installation/azure.md
```diff
@@ -81,6 +81,7 @@ Update the `appsettings.json` file:
         "Type": "Azure",
         "AccountName": "my-account",
         "ApiKey": "ABCD1234"
+        "PackageIdFilter": "^BlacklistRegex$"
```
loic-sharma, line 84: Please use terminology like "block list" instead

DavidHamburg:
> Hi, thanks for opening this pull request! Wouldn't it be better to filter results in the [mirror service](https://github.com/loic-sharma/BaGet/blob/master/src/BaGet.Core/Mirror/MirrorService.cs), which is what discovers packages from an upstream NuGet server like nuget.org?

Yes, indeed. I have moved the implementation to the Mirror-Service.

> Also, what do you think of supporting wildcards instead of regex? For example, if you use _xyz_ as your package ID prefix, your filter could be `xyz.*`

The regular expressions are more flexible than wildcard filters. One case I have in mind is to negate the search term to use it as positiv filter.

> If you move the filter to the mirror service, what do you think of a configuration like:
> 
> ```json
> {
>   "Mirror": {
>     "Enabled": true,
>     "PackageSource": "https://api.nuget.org/v3/index.json",
>     "Exclude": [
>       "xyz.*"
>      ]
>   }
> }
> ```

Thanks for the hint, I have updated the key to ```Exclude```.

PR #562, opened 2020-07-23
[Search] Refactor implementation

loic-sharma:
Improves the search services:

1. Introduces request models to reduce number of parameters
1. Plumbs additional search parameters to the autocomplete endpoint
1. Adds `packageType` property to the search response model
1. Removes unneeded default values on search services


src/BaGet.Core/Search/ISearchService.cs
```diff
@@ -25,48 +25,57 @@ public interface ISearchService
...
-            int take = 20,
-            bool includePrerelease = true,
-            bool includeSemVer2 = true,
-            CancellationToken cancellationToken = default);
+            string query,
+            int skip,
+            int take,
+            bool includePrerelease,
+            bool includeSemVer2,
+            CancellationToken cancellationToken);
```
loic-sharma, line 54: TODO: Add package type filter

src/BaGet.Hosting/Controllers/SearchController.cs
```diff
@@ -58,9 +62,7 @@ public SearchController(ISearchService searchService)
             CancellationToken cancellationToken = default)
         {
             // TODO: Add other dependents parameters.
```
loic-sharma, line 64: Clean this

Issue #563, opened 2020-07-23
docker image dont run in 5000 run in 80


behroozbc:
### Describe the bug
hi 
i want install baget form docker and see in document port 5000 but baget run in port 80 

![image](https://user-images.githubusercontent.com/20381616/88279563-89db6000-ccf9-11ea-859a-5488daacef16.png)

ashishbhatia22:
We should modify the docker file and expose port 5000. Currently, only port 80 is exposed

behroozbc:
@ashishbhatia22
Hi
Don't think when I check logs I see baget run on port 80 and don't run in 5000
In asp. Net core port 5000 for test and development

abakumov-v:
Hello! Same issue. May be just change BaGet docs at this page https://loic-sharma.github.io/BaGet/installation/docker/ ?

loic-sharma:
Thank you for reporting this! @behroozbc fixed this in https://github.com/loic-sharma/BaGet/pull/564

PR #564, opened 2020-07-23
change port 5000 to 80

behroozbc:
hi 
i want install baGet and  I see run in port 80 not 5000 
I update docker run command

change docker port as 5000 to 80 
https://github.com/loic-sharma/BaGet/issues/563


loic-sharma:
Thanks for fixing this!

PR #573, opened 2020-08-17
Fix package filtering with TableSearchService

yang-er:
Support filtering TargetFrameworks and PackageType in BaGet.Azure.TableSearchService in backend service.

### DIFF
--- a/src/BaGet.Azure/Table/TableSearchService.cs
+++ b/src/BaGet.Azure/Table/TableSearchService.cs
@@ -47,6 +47,30 @@ namespace BaGet.Azure
                 request.IncludeSemVer2,
                 cancellationToken);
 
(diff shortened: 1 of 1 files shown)


loic-sharma:
Hi, thanks for sending this contribution! Could you explain a little more why you'd like to add this? This won't work as expected with the `skip` and `take` parameters. Support all search features using the Azure Table search service is extremely difficult, I was planning on supporting additional features. Happy to hear your thoughts though! :)

yang-er:
Currently if we couldn't implement the searching well with Azure Tables, we have two choices:

- To show correct search results, but each page loses several items.
- To show correct numbers of items, but the search result is not filtered.

I think the first choice is much better than the second one. The first one may mislead users but the second one only requires users to navigate more pages to see the results.

PR #574, opened 2020-08-17
Fix multiple author displaying

yang-er:
Previously, the author is displayed as `by: author1author2`. This fix make it `by: author1 author2`.


loic-sharma:
Thanks for the contribution, and nice catch! It looks like you'll need to update this test case to return an array of strings:

https://github.com/loic-sharma/BaGet/blob/7c514bfe57a59b7ed086e37b73d61a8194306a20/src/BaGet.UI/src/SearchResults.test.tsx#L38

loic-sharma:
Thanks for the contribution!

PR #575, opened 2020-08-22
Integration Tests for Dependents API

davidackroyd99:
Tests that the dependents api:

* Returns OK when packageId=DefaultPackage
* Returns BadRequest when a package ID is not supplied (I implemented this as well, I think it makes sense TBH)

These are my first integration tests ever so let me know how I could improve! I want to add tests for the icon and readme endpoints but I guess you probably want to approve these first before I do that.

### DIFF
--- a/src/BaGet.Hosting/Controllers/SearchController.cs
+++ b/src/BaGet.Hosting/Controllers/SearchController.cs
@@ -69,9 +69,14 @@ namespace BaGet.Hosting
         }
 
         public async Task<ActionResult<DependentsResponse>> DependentsAsync(
-            [FromQuery] string packageId,
+            [FromQuery] string packageId = null,
             CancellationToken cancellationToken = default)
         {
+            if(string.IsNullOrWhiteSpace(packageId))
+            {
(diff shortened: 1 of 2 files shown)


src/BaGet.Hosting/Controllers/SearchController.cs
```diff
@@ -69,9 +69,14 @@ public SearchController(ISearchService searchService)
         }
 
         public async Task<ActionResult<DependentsResponse>> DependentsAsync(
-            [FromQuery] string packageId,
+            [FromQuery] string packageId = null,
             CancellationToken cancellationToken = default)
         {
+            if(string.IsNullOrWhiteSpace(packageId))
```
loic-sharma, line 75:
```suggestion
            if (string.IsNullOrWhiteSpace(packageId))
```

loic-sharma:
Thanks for adding these tests!

> I implemented this as well, I think it makes sense TBH

Yup, agreed.

> These are my first integration tests ever so let me know how I could improve! 

These are perfect, nice work 😎

Issue #291, opened 2019-07-03
Support version autocomplete


loic-sharma:
# Work

* [x] Database support https://github.com/loic-sharma/BaGet/pull/576
* [ ] Azure Search support
* [ ] Azure Tables support

# Background

Implement this protocol: https://docs.microsoft.com/en-us/nuget/api/search-autocomplete-service-resource#enumerate-package-versions

Here's an example from nuget.org: https://azuresearch-usnc.nuget.org/autocomplete?id=newtonsoft.json&prerelease=true&semVerLevel=2.0.0

You can find places that need to be updated here: https://github.com/loic-sharma/BaGet/search?q=https%3A%2F%2Fgithub.com%2Floic-sharma%2FBaGet%2Fissues%2F291&unscoped_q=https%3A%2F%2Fgithub.com%2Floic-sharma%2FBaGet%2Fissues%2F291

loic-sharma:
Thanks @davidackroyd99 for implementing this feature!

loic-sharma:
Actually, I'll reopen this as we still need to add Azure Search and Azure Tables support

PR #576, opened 2020-08-23
Support for version autocomplete

davidackroyd99:
Does not allow version search support for Azure yet - I have no way of testing this, when this PR is approved as being in roughly the right direction I'll try and set up a test env.

Addresses https://github.com/loic-sharma/BaGet/issues/291

### DIFF
--- a/src/BaGet.Core/Search/DatabaseSearchService.cs
+++ b/src/BaGet.Core/Search/DatabaseSearchService.cs
@@ -129,8 +129,6 @@ namespace BaGet.Core
                 .Select(p => p.Version.ToString())
                 .ToListAsync(cancellationToken);
 
-            results.Sort();
-
             return new AutocompleteResponse
             {
                 TotalHits = results.Count,
--- a/src/BaGet.Core/Search/DatabaseSearchService.cs
+++ b/src/BaGet.Core/Search/DatabaseSearchService.cs
@@ -106,13 +106,38 @@ namespace BaGet.Core
             };
         }
 
-        public Task<AutocompleteResponse> ListPackageVersionsAsync(
+        public async Task<AutocompleteResponse> ListPackageVersionsAsync(
             AutocompleteRequest request,
             CancellationToken cancellationToken)
         {
-            // TODO: Support versions autocomplete.
-            // See: https://github.com/loic-sharma/BaGet/issues/291
-            throw new NotImplementedException();
+            IQueryable<Package> search = _context.Packages;
+
+            if (!string.IsNullOrEmpty(request.Query))
+            {
+                var query = request.Query.ToLower();
+                search = search.Where(p => p.Id.ToLower().Equals(query));
+            }
+
+            search = AddSearchFilters(
+                search,
+                request.IncludePrerelease,
+                request.IncludeSemVer2,
+                request.PackageType,
+                frameworks: null);
+
+            var results = await search
+                .Distinct()
+                .Select(p => p.Version.ToString())
+                .ToListAsync(cancellationToken);
+
+            results.Sort();
+
+            return new AutocompleteResponse
+            {
+                TotalHits = results.Count,
+                Data = results,
+                Context = AutocompleteContext.Default
+            };
         }
 
         public async Task<DependentsResponse> FindDependentsAsync(string packageId, CancellationToken cancellationToken)
--- a/src/BaGet.Core/Search/DatabaseSearchService.cs
+++ b/src/BaGet.Core/Search/DatabaseSearchService.cs
@@ -110,23 +110,20 @@ namespace BaGet.Core
             VersionsRequest request,
             CancellationToken cancellationToken)
         {
-            IQueryable<Package> search = _context.Packages;
-
-            if (!string.IsNullOrEmpty(request.Query))
-            {
-                var query = request.Query.ToLower();
-                search = search.Where(p => p.Id.ToLower().Equals(query));
-            }
+            var packageId = request.PackageId.ToLower();
+            IQueryable<Package> search = _context
+                .Packages
+                .Where(p => p.Id.ToLower().Equals(packageId));
 
             search = AddSearchFilters(
                 search,
                 request.IncludePrerelease,
                 request.IncludeSemVer2,
-                request.PackageType,
+                packageType: null,
                 frameworks: null);
 
             var results = await search
-                .Select(p => p.Version.ToString())
+                .Select(p => p.Version.ToNormalizedString())
                 .ToListAsync(cancellationToken);
 
             return new AutocompleteResponse
--- a/src/BaGet.Core/Search/DatabaseSearchService.cs
+++ b/src/BaGet.Core/Search/DatabaseSearchService.cs
@@ -107,7 +107,7 @@ namespace BaGet.Core
         }
 
         public async Task<AutocompleteResponse> ListPackageVersionsAsync(
-            AutocompleteRequest request,
+            VersionsRequest request,
             CancellationToken cancellationToken)
         {
             IQueryable<Package> search = _context.Packages;
--- a/src/BaGet.Core/Search/DatabaseSearchService.cs
+++ b/src/BaGet.Core/Search/DatabaseSearchService.cs
@@ -126,7 +126,6 @@ namespace BaGet.Core
                 frameworks: null);
 
             var results = await search
-                .Distinct()
                 .Select(p => p.Version.ToString())
                 .ToListAsync(cancellationToken);
 
--- a/src/BaGet.Core/Search/DatabaseSearchService.cs
+++ b/src/BaGet.Core/Search/DatabaseSearchService.cs
@@ -107,9 +107,7 @@ namespace BaGet.Core
         }
 
         public Task<AutocompleteResponse> ListPackageVersionsAsync(
-            string packageId,
-            bool includePrerelease,
-            bool includeSemVer2,
+            AutocompleteRequest request,
             CancellationToken cancellationToken)
         {
             // TODO: Support versions autocomplete.
--- a/src/BaGet.Core/Search/DatabaseSearchService.cs
+++ b/src/BaGet.Core/Search/DatabaseSearchService.cs
@@ -106,7 +106,7 @@ namespace BaGet.Core
             };
         }
 
-        public Task<AutocompleteResponse> ListPackageVersionsAssync(
+        public Task<AutocompleteResponse> ListPackageVersionsAsync(
             string packageId,
             bool includePrerelease,
             bool includeSemVer2,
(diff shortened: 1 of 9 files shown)


src/BaGet.Hosting/Controllers/SearchController.cs
```diff
@@ -43,29 +43,42 @@ public SearchController(ISearchService searchService)
...
+            }
+
+            if(versionsQuery != null)
+            {
+                request.Query = versionsQuery;
+
+                return await _searchService.ListPackageVersionsAsync(request, cancellationToken);
+            }
+
+            return new AutocompleteResponse();
```
loic-sharma, line 81: Your logic is perfectly correct but I'd prefer to rearrange it so that line 81 isn't necessary. What do you think of something like:

```suggestion
            // If only "id" is provided, find package versions. Otherwise, find package ids.
            if (versionsQuery != null && autocompleteQuery == null)
            {
                request.Query = versionsQuery;

                return await _searchService.ListPackageVersionsAsync(request, cancellationToken);
            }
            else
            {
                request.Query = autocompleteQuery;

                return await _searchService.AutocompleteAsync(request, cancellationToken);
            }
```

src/BaGet.Core/Search/ISearchService.cs
```diff
@@ -33,16 +33,10 @@ public interface ISearchService
...
-        /// <param name="includeSemVer2">Whether packages that require SemVer 2.0.0 compatibility should be returned.</param>
+        /// <param name="request">The autocomplete request.</param>
         /// <param name="cancellationToken">A token to cancel the task.</param>
         /// <returns>The package versions that matched the request.</returns>
-        Task<AutocompleteResponse> ListPackageVersionsAssync(
-            string packageId,
-            bool includePrerelease,
-            bool includeSemVer2,
-            CancellationToken cancellationToken);
+        Task<AutocompleteResponse> ListPackageVersionsAsync(AutocompleteRequest request, CancellationToken cancellationToken);
```
loic-sharma, line 39: The list package versions endpoint will always list *all* versions and doesn't accept `skip` or `take` parameters so I don't think we should reuse the `AutocompleteRequest` type here.

For more information, see these docs: https://docs.microsoft.com/en-us/nuget/api/search-autocomplete-service-resource#enumerate-package-versions

src/BaGet.Hosting/Controllers/SearchController.cs
```diff
@@ -43,29 +43,42 @@ public SearchController(ISearchService searchService)
         }
 
         public async Task<ActionResult<AutocompleteResponse>> AutocompleteAsync(
-            [FromQuery(Name = "q")] string query = null,
-            [FromQuery]int skip = 0,
-            [FromQuery]int take = 20,
+            [FromQuery(Name = "q")] string autocompleteQuery = null,
+            [FromQuery(Name = "id")] string versionsQuery = null,
```
loic-sharma, line 47: I like these names 👍

src/BaGet.Core/Search/DatabaseSearchService.cs
```diff
@@ -106,15 +106,38 @@ public DatabaseSearchService(IContext context, IFrameworkCompatibilityService fr
...
+                request.IncludeSemVer2,
+                request.PackageType,
+                frameworks: null);
+
+            var results = await search
+                .Distinct()
+                .Select(p => p.Version.ToString())
+                .ToListAsync(cancellationToken);
+
+            results.Sort();
```
loic-sharma, line 133: The results don't need to be sorted, so let's remove this line. (Also, string sorting isn't quite right here as it yields a different order than semver sorting. To get the desired result we'd have to parse all the versions before sorting them)

src/BaGet.Core/Search/DatabaseSearchService.cs
```diff
@@ -106,15 +106,38 @@ public DatabaseSearchService(IContext context, IFrameworkCompatibilityService fr
...
+
+            search = AddSearchFilters(
+                search,
+                request.IncludePrerelease,
+                request.IncludeSemVer2,
+                request.PackageType,
+                frameworks: null);
+
+            var results = await search
+                .Distinct()
```
loic-sharma, line 129: Is the `.Distinct()` here necessary? I think we can remove this

loic-sharma:
This looks great, thanks for doing this!

> ... when this PR is approved as being in roughly the right direction ...

Yup this looks like the right direction. I left a few minor nitpicks, my only real feedback is that we shouldn't reuse the `AutocompleteRequest` for the `ListPackageVersionsAsync` method as they have separate request parameters.

Please let me know if you have any questions with the Azure Search implementation! This may be a good opportunity for me to beef up our developer docs 😊

src/BaGet.Hosting/Controllers/SearchController.cs
davidackroyd99: cheers man 👍

src/BaGet.Core/Search/ISearchService.cs
davidackroyd99: I've made a separate VersionsRequest object, I know that there is quite a lot of code duplication inherent in doing that but it felt like it made it the easiest to understand

src/BaGet.Core/Search/DatabaseSearchService.cs
```diff
@@ -106,15 +106,35 @@ public DatabaseSearchService(IContext context, IFrameworkCompatibilityService fr
...
+
+            search = AddSearchFilters(
+                search,
+                request.IncludePrerelease,
+                request.IncludeSemVer2,
+                request.PackageType,
+                frameworks: null);
+
+            var results = await search
+                .Select(p => p.Version.ToString())
```
loic-sharma, line 129: You can use [`NormalizedVersionString`](https://github.com/loic-sharma/BaGet/blob/f0ee2c57c221fd6f0e866f051eb5892e464b85f0/src/BaGet.Core/Entities/Package.cs#L67) directly instead (the `Version` property has to parse the version string)

```suggestion
                .Select(p => p.NormalizedVersionString)
```

src/BaGet.Core/Search/VersionsRequest.cs
```diff
@@ -0,0 +1,29 @@
+namespace BaGet.Core
+{
+    /// <summary>
+    /// The NuGet V3 search request.
+    /// See: https://docs.microsoft.com/en-us/nuget/api/search-autocomplete-service-resource#request-parameters
```
loic-sharma, line 5:
```suggestion
    /// The NuGet V3 enumerage package versions request.
    /// See: https://docs.microsoft.com/en-us/nuget/api/search-autocomplete-service-resource#request-parameters-1
```

src/BaGet.Core/Search/VersionsRequest.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        /// <summary>
+        /// Whether to include SemVer 2.0.0 compatible packages.
+        /// </summary>
+        public bool IncludeSemVer2 { get; set; }
+
+        /// <summary>
+        /// Filter results to a package type. If null, no filter is applied.
+        /// </summary>
+        public string PackageType { get; set; }
```
loic-sharma, line 22: Please remove the `PackageType` property as the autocomplete package versions API doesn't accept a package type filter (see [this](https://docs.microsoft.com/en-us/nuget/api/search-autocomplete-service-resource#request-parameters-1)).

src/BaGet.Core/Search/VersionsRequest.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        /// <summary>
+        /// The search query.
+        /// </summary>
+        public string Query { get; set; }
```
loic-sharma, line 27: What do you think of renaming this property to `PackageId`?

```suggestion
        /// <summary>
        /// The package ID whose versions should be fetched.
        /// </summary>
        public string PackageId { get; set; }
```

src/BaGet.Core/Search/DatabaseSearchService.cs
```diff
@@ -106,15 +106,35 @@ public DatabaseSearchService(IContext context, IFrameworkCompatibilityService fr
...
+            {
+                var query = request.Query.ToLower();
+                search = search.Where(p => p.Id.ToLower().Equals(query));
+            }
+
+            search = AddSearchFilters(
+                search,
+                request.IncludePrerelease,
+                request.IncludeSemVer2,
+                request.PackageType,
```
loic-sharma, line 125: The list package versions API doesn't support package type filtering, let's just pass `null` here:

```suggestion
                packageType: null,
```

src/BaGet.Core/Search/DatabaseSearchService.cs
```diff
@@ -106,15 +106,35 @@ public DatabaseSearchService(IContext context, IFrameworkCompatibilityService fr
...
-            // TODO: Support versions autocomplete.
-            // See: https://github.com/loic-sharma/BaGet/issues/291
-            throw new NotImplementedException();
+            IQueryable<Package> search = _context.Packages;
+
+            if (!string.IsNullOrEmpty(request.Query))
+            {
+                var query = request.Query.ToLower();
+                search = search.Where(p => p.Id.ToLower().Equals(query));
+            }
```
loic-sharma, line 119: We can simplify this a little since the `id` query parameter is required for this API:

```suggestion
            var packageId = request.Query.ToLower();
            IQueryable<Package> search = _context
                .Packages
                .Where(p => p.Id.ToLower().Equals(query));
```

loic-sharma:
> Does not allow version search support for Azure yet...

On second thought, I think it'd be fine for this pull request to support just the database. We can add Azure Search support later in a separate pull request.

davidackroyd99:
> We can add Azure Search support later in a separate pull request.

Good idea, I've made a few more little changes and it passes the tests so if we get this in I can revisit for the Azure stuff, after we've written some integration tests for it.

src/BaGet.Core/Search/DatabaseSearchService.cs
```diff
@@ -106,15 +106,32 @@ public DatabaseSearchService(IContext context, IFrameworkCompatibilityService fr
...
+
+            search = AddSearchFilters(
+                search,
+                request.IncludePrerelease,
+                request.IncludeSemVer2,
+                packageType: null,
+                frameworks: null);
+
+            var results = await search
+                .Select(p => p.Version.ToNormalizedString())
```
loic-sharma, line 126: Please avoid using the `p.Version` property as it actually parses the `OriginalVersionString` property. Also, I'm not sure this plays nicely with EF Core, it may require client side evaluation (as the `ToNormalizedString` method can't be translated to SQL). Please use the `p.NormalizedVersionString` property instead:

```suggestion
                .Select(p => p.NormalizedVersionString)
```

src/BaGet.Core/Search/VersionsRequest.cs
```diff
@@ -0,0 +1,24 @@
+namespace BaGet.Core
+{
+    /// <summary>
+    /// The NuGet V3 enumerage package versions request.
```
loic-sharma, line 4:
```suggestion
    /// The NuGet V3 enumerate package versions request.
```

loic-sharma:
Thanks for your contribution!

Issue #97, opened 2018-09-10
GitHub Integration for repository URL


loic-sharma:
The number of open pull requests, open issues, and the last commit date should be fetched if the package has a repository URL.

These API calls should be kicked off by the client and should be unauthenticated. If the rate limit of 60 requests per hour is hit, the client should gracefully hide the GitHub details.

See the Visual Studio Marketplace: https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-pull-request-github

Use GitHub release information for changelog?

PR #578, opened 2020-08-24
GitHub Integration (Proof of Concept)

davidackroyd99:
I did this, but the GitHub API often seems to return incorrect results for open issues (whether you use 'open_issues' or 'open_issues_count'), and doesn't return number of open PRs at all. Would we not be better off listing the number of stars and forks?

Once we have decided on stats and so on to return I will clean up the code, but this works as a proof of concept.

Addresses https://github.com/loic-sharma/BaGet/issues/97

### DIFF
--- a/src/BaGet.UI/src/DisplayPackage/DisplayPackage.tsx
+++ b/src/BaGet.UI/src/DisplayPackage/DisplayPackage.tsx
@@ -315,9 +315,7 @@ class DisplayPackage extends React.Component<IDisplayPackageProps, IDisplayPacka
               <p>{(!this.state.package.authors) ? 'Unknown' : this.state.package.authors}</p>
             </div>
 
-            <div>
-              <RepoDetails url={this.state.package.projectUrl}/>
-            </div>
+            <RepoDetails url={this.state.package.projectUrl}/>
           </aside>
         </div>
       );
--- a/src/BaGet.UI/src/DisplayPackage/DisplayPackage.tsx
+++ b/src/BaGet.UI/src/DisplayPackage/DisplayPackage.tsx
@@ -11,6 +11,7 @@ import { PackageType, InstallationInfo } from './InstallationInfo';
 import LicenseInfo from './LicenseInfo';
 import * as Registration from './Registration';
 import SourceRepository from './SourceRepository';
+import RepoDetails from './RepoDetails';
 import { Versions, IPackageVersion } from './Versions';
 
 import './DisplayPackage.css';
@@ -312,6 +313,10 @@ class DisplayPackage extends React.Component<IDisplayPackageProps, IDisplayPacka
               <h2>Authors</h2>
 
               <p>{(!this.state.package.authors) ? 'Unknown' : this.state.package.authors}</p>
+            </div>
+
+            <div>
+              <RepoDetails />
             </div>
           </aside>
         </div>
--- a/src/BaGet.UI/src/DisplayPackage/DisplayPackage.tsx
+++ b/src/BaGet.UI/src/DisplayPackage/DisplayPackage.tsx
@@ -316,7 +316,7 @@ class DisplayPackage extends React.Component<IDisplayPackageProps, IDisplayPacka
             </div>
 
             <div>
-              <RepoDetails />
+              <RepoDetails url={this.state.package.projectUrl}/>
             </div>
           </aside>
         </div>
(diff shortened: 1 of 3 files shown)


loic-sharma:
!!!! This is awesome. I'll review this in a few hours when I'm off work

loic-sharma:
>  the GitHub API often seems to return incorrect results for open issues (whether you use 'open_issues' or 'open_issues_count'), and doesn't return number of open PRs at all. Would we not be better off listing the number of stars and forks?

I think showing the number of stars and forks is a great stars. Personally I'm not too interested in the number of issues or PRs.

src/BaGet.UI/src/DisplayPackage/DisplayPackage.tsx
```diff
@@ -313,6 +314,8 @@ class DisplayPackage extends React.Component<IDisplayPackageProps, IDisplayPacka
 
               <p>{(!this.state.package.authors) ? 'Unknown' : this.state.package.authors}</p>
             </div>
+
+            <RepoDetails url={this.state.package.projectUrl}/>
```
loic-sharma, line 318: The `ICatalogEntry` I made is incorrect. The `projectUrl` property should be nullable. See: https://github.com/loic-sharma/BaGet/blob/f0ee2c57c221fd6f0e866f051eb5892e464b85f0/src/BaGet.UI/src/DisplayPackage/Registration.tsx#L26

src/BaGet.UI/src/DisplayPackage/RepoDetails.tsx
```diff
@@ -0,0 +1,105 @@
...
+}
+
+interface IRepoDetailsState {
+  openIssues: number;
+  openPulls: number;
+  lastCommit: Date;
+  detailsFound: boolean;
+}
+
+class RepoDetails extends React.Component<IRepoDetailsProps, IRepoDetailsState> {
```
loic-sharma, line 18: Just curious, why did you create a new component instead of updating the [`SourceRepository`](https://github.com/loic-sharma/BaGet/blob/master/src/BaGet.UI/src/DisplayPackage/SourceRepository.tsx) component?

src/BaGet.UI/src/DisplayPackage/RepoDetails.tsx
```diff
@@ -0,0 +1,105 @@
...
+    this.state = {
+      openIssues: 0,
+      openPulls: 0,
+      lastCommit: new Date(),
+      detailsFound: false,
+    };
+  }
+
+  private getRepoOwnerAndName(repoUrl: string) {
+    const frags = repoUrl.split("/");
```
loic-sharma, line 36: We should check that the repository URL's hostname is GitHub. You should be able to use something like:

```ts
if (new URL(repoUrl).hostname === 'github.com')
```

See: https://developer.mozilla.org/en-US/docs/Web/API/URL/hostname

src/BaGet.UI/src/DisplayPackage/RepoDetails.tsx
```diff
@@ -0,0 +1,105 @@
...
+        }).then(json => {
+          console.log(json);
+          this.setState({
+            detailsFound: true,
+            openIssues: json['open_issues'],
+            openPulls: 0,
+            lastCommit: new Date(json['pushed_at'])
+          });
+        // tslint:disable-next-line:no-console
+        }).catch((e) => console.log("Failed to load dependents.", e));
```
loic-sharma, line 73: I believe the GitHub API will rate limit us very quickly. We should handle this gracefully and not display any information.

src/BaGet.UI/src/DisplayPackage/RepoDetails.tsx
loic-sharma: Also, you can split on the url's `pathname`: https://developer.mozilla.org/en-US/docs/Web/API/URL/pathname

src/BaGet.UI/src/DisplayPackage/RepoDetails.tsx
davidackroyd99: Once I've finished this I might remove the heading tag and put it inside the SourceRepository component, this is a good point

src/BaGet.UI/src/DisplayPackage/RepoDetails.tsx
davidackroyd99: Cheers for this, will use it 👍

src/BaGet.UI/src/DisplayPackage/RepoDetails.tsx
davidackroyd99: We do, because the foundDetails state will be set to false so it won't render

src/BaGet.UI/src/DisplayPackage/DisplayPackage.tsx
davidackroyd99: Will do!

PR #585, opened 2020-09-02
Create pushing_packages.md

akovac35:
Added a more elaborate example on how to push packages.

### DIFF
--- a/docs/advanced/pushing_packages.md
+++ b/docs/advanced/pushing_packages.md
@@ -1,0 +1,70 @@
+# Pushing packages to BaGet
+
+The following **bash** script can serve as a reference when pushing **nupkg** and **snupkg** solution project packages to BaGet - there can be more than one project in a solution and this script will push all of them:
+
+```sh
+# Authors
+#	Aleksander Kovač https://github.com/akovac35
+#
+# This script will push all Release project packages for a solution to <BaGet host>:<BaGet port>.
+#
+# Specific solution projects can be excluded with the following setting:
+#  <PropertyGroup>
+#    <IsPackable>false</IsPackable>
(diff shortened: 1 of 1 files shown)


loic-sharma:
Hi, thanks for opening this pull request. I'm concerned that this script is a little too specific for most users. The `*` wildcard feature on `dotnet nuget push` is probably enough for most people. What are your thoughts?

akovac35:
> Hi, thanks for opening this pull request. I'm concerned that this script is a little too specific for most users. The `*` wildcard feature on `dotnet nuget push` is probably enough for most people. What are your thoughts?

Well the snippet is definitively useful because it deals in one place with many of the typical problems encountered when pushing nugets:

- detect a test has failed,
- clean, build, pack "Release" variants of all solution projects,
- find the above and push to BaGet,
- configuration steps for uploading nugets when the repository is protected with a key,
- just gives a general idea and a starting point for many common nuget related matters, such as how to skip packing a project and creating and pushing symbol packages which is typically an afterthought.

```dotnet nuget push``` definitively does not cut it most of the time, or I would not go through the hassle. For example, it might push debug variants if we are not careful, then there is the hassle with the key, which you have to find online, and just don't get me started on the star *, I wasn't even aware it exists.

To conclude, BaGet will get loads more stars just because of this script ;)

loic-sharma:
Your script is definitely useful, but I don't think it's a good fit for BaGet's documentation. I'd prefer to document the basics geared towards beginners. For example, how do I upload my first package? How do I import packages from some other server? How do I use MySQL?

Your script is more for advanced users that are looking for best practices. This would make for a great blog post though!

akovac35:
I will make a gist instead.

Issue #594, opened 2020-09-29
Out-Of-Memory error when pushing 200 MB package with S3 storage


RoguePointer80:
### Describe the bug
Our Baget instance is running in a Kubernetes environment with 1 GB or RAM requested and 2 GB Max limit configured.
The storage backend is S3. 
One of our teams has a big package which is 200 MB. When adding the package to nuget, the there is an out-of-memory error. Here is the callstack:
```
info: Microsoft.Hosting.Lifetime[0]
      Now listening on: http://[::]:80
info: Microsoft.Hosting.Lifetime[0]
      Application started. Press Ctrl+C to shut down.
info: Microsoft.Hosting.Lifetime[0]
      Hosting environment: Production
info: Microsoft.Hosting.Lifetime[0]
      Content root path: /app
fail: BaGet.Core.PackageIndexingService[0]
      Failed to persist package Coveo.Connectors.OnPremComponent 8.0.5366.3 content to storage
System.OutOfMemoryException: Exception of type 'System.OutOfMemoryException' was thrown.
   at System.IO.MemoryStream.set_Capacity(Int32 value)
   at System.IO.MemoryStream.EnsureCapacity(Int32 value)
   at System.IO.MemoryStream.Write(Byte[] buffer, Int32 offset, Int32 count)
   at System.IO.MemoryStream.WriteAsync(ReadOnlyMemory`1 buffer, CancellationToken cancellationToken)
--- End of stack trace from previous location where exception was thrown ---
   at System.IO.Stream.CopyToAsyncInternal(Stream destination, Int32 bufferSize, CancellationToken cancellationToken)
   at BaGet.Aws.S3StorageService.PutAsync(String path, Stream content, String contentType, CancellationToken cancellationToken) in /src/BaGet.Aws/S3StorageService.cs:line 81
   at BaGet.Core.PackageStorageService.SavePackageContentAsync(Package package, Stream packageStream, Stream nuspecStream, Stream readmeStream, Stream iconStream, CancellationToken cancellationToken) in /src/BaGet.Core/Storage/PackageStorageService.cs:line 58
   at BaGet.Core.PackageIndexingService.IndexAsync(Stream packageStream, CancellationToken cancellationToken) in /src/BaGet.Core/Indexing/PackageIndexingService.cs:line 105
fail: BaGet.Hosting.PackagePublishController[0]
      Exception thrown during package upload
System.OutOfMemoryException: Exception of type 'System.OutOfMemoryException' was thrown.
[...]

PR #595, opened 2020-10-01
Remove unnecessary stream copy in AWS S3 Put

RoguePointer80:
Remove unnecessary copy of package stream during AWS S3 PutAsync

 * Tested locally (local docker, but using S3 storage), with the same offending 200 MB package as reported in the issue.

Addresses https://github.com/loic-sharma/BaGet/issues/594

I ran `docker stats baget` to check the memory usage of the process, and I got
```
CONTAINER ID        NAME                CPU %               MEM USAGE / LIMIT   MEM %               NET I/O             BLOCK I/O           PIDS
9c57eea5cccc        baget               0.13%               147.8MiB / 1GiB     14.43%              204MB / 206MB       0B / 0B             25
```
So we see that the memory usage went up to only 150 MB, which is quite ok.

### DIFF
--- a/src/BaGet.Aws/S3StorageService.cs
+++ b/src/BaGet.Aws/S3StorageService.cs
@@ -74,22 +74,15 @@ namespace BaGet.Aws
             // TODO: Uploads should be idempotent. This should fail if and only if the blob
             // already exists but has different content.
 
-            using (var seekableContent = new MemoryStream())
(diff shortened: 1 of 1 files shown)


RoguePointer80:
@loic-sharma   This basic fix has been idling for a long time. Could you please review it?

Issue #75, opened 2018-08-05
Health Endpoint Monitoring


WernerMairl:
Feature/Pattern

https://docs.microsoft.com/en-us/azure/architecture/patterns/health-endpoint-monitoring

at least in the simplest possible way....

WernerMairl:
foun some interesting project:

Prototype for ASP.NETCore 2.2: https://github.com/dotnet-architecture/HealthChecks
https://github.com/Xabaril/BeatPulse

WernerMairl:
i suggest 
- one health Get-request for the whole app (implementing a HealthController that does the work)

- one (smaller) Health-get-Method on each Controller

Usecase for "one health check per Controller)

we had false positive results in the past testing single controllers 
because we have tested against not existing Packages expecting HttpReturncode.NotFound

the controller was not starting on TestHost... TestHost returns HttpReturncode.NotFound
=> two different usecases, one returncode....
if we can do a health check (=check if Controller is running) that may help in situations like that

RoguePointer80:
I just made a PR for adding basic health checks. I would think that it fills your "one health request for the whole app" suggestion, although it's really two.
I understand your issue about the controller that was not started; I'm not sure how best to implement checking that a controller is loaded. The way I did my health checks is to have a health check class that get the service implementation given by the DI, like IStorageService. I could do the same for IPackageContentService for example; but that wouldn't catch the case where the PackageContentController is not started even though the the service is available in the DI container.

The use-case I am trying to solve with my basic health checks relates first to a bad configuration: maybe the database connection string is incorrect; or maybe the storage parameters point to a read-only directory, or that the Azure/AWS credentials are not found. Any of these configuration errors would make the health check fail. Secondly, another use-case is when the availability of these external resources changes. For example maybe the database went offline for an unexpected upgrade; in that case the health check would suddenly start failing.

Could you please share here what kind of error would cause a controller not starting? Is it something that can be catched with a unit test or integration test?

PR #596, opened 2020-10-02
Feature/healthcheck3

RoguePointer80:
This PR adds a health check endpoint with basic functionality

 * Database check is simply "can connect"
 * Storage check does a Put, Get and Delete operations

Addresses https://github.com/loic-sharma/BaGet/issues/75


RoguePointer80:
I don't really understand why my last commit was necessary. I debugged the unit test, and it appears that the Startup class ConfigureServices is called twice when creating the unit test.
I found this issue :  https://github.com/dotnet/aspnetcore/issues/20307  which is reporting exactly the behaviour I see, however I don't understand the recommendation (use Mocks?)... well I do know what mocks are, I just don't see how that would prevent our startup class being called twice.

PR #597, opened 2020-10-05
Explain how to run BaGet on AWS.

monty241:
Describe that BaGet requires Docker or alike to execute the server process in combination with storage on database and S3.


BlythMeister (approved):
LGTM

Issue #590, opened 2020-09-08
Change ReleaseNotes to nvarchar(max) in DB schema for SqlServer


gimlichael:
### Describe the bug
When uploading a NuGet package with a thorough release notes, an SqLException is thrown.
According to the specs of NuGet, 35.000 characters is allowed.
I recommend nvarchar(max) given the current design of your schema.

### To Reproduce
1. Docker image: loicsharma/baget:19dc5c514fcbfdeda0c06d67b526744733d827bb
2. Upload a NuGet package with long release notes

### Expected behavior
I would expect BaGet to conform with NuGet specifications: https://docs.microsoft.com/en-us/nuget/reference/nuspec

### Exception
```
2020-09-08T19:40:54.383725267Z fail: Microsoft.EntityFrameworkCore.Update[10000]
2020-09-08T19:40:54.383785367Z       An exception occurred in the database while saving changes for context type 'BaGet.Database.SqlServer.SqlServerContext'.
2020-09-08T19:40:54.383801268Z       Microsoft.EntityFrameworkCore.DbUpdateException: An error occurred while updating the entries. See the inner exception for details.
2020-09-08T19:40:54.383805368Z        ---> Microsoft.Data.SqlClient.SqlException (0x80131904): String or binary data would be truncated in table 'devnuget.dbo.Packages', column 'ReleaseNotes'. Truncated value: 'Version: 6.0.0
2020-09-08T19:40:54.383808868Z       Availability: NET Standard 2.0
2020-09-08T19:40:54.383811768Z        
2020-09-08T19:40:54.383814968Z       # Upgrade Steps
2020-09-08T19:40:54.383817868Z       - The Cuemon.Serialization.Xml names'.
2020-09-08T19:40:54.383820768Z       The statement has been terminated.
2020-09-08T19:40:54.383823668Z          at Microsoft.Data.SqlClient.SqlCommand.<>c.<ExecuteDbDataReaderAsync>b__164_0(Task`1 result)
2020-09-08T19:40:54.383827268Z          at System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke()
2020-09-08T19:40:54.383829968Z          at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
2020-09-08T19:40:54.383832868Z       --- End of stack trace from previous location where exception was thrown ---
[...]

Issue #609, opened 2020-11-17
Baget unable to mirror package (Microsoft.Data.SqlClient.SqlException (0x80131904): String or binary data would be truncated.)


Svisstack:
### Describe the bug

Baget is unable to cache a package on Sql Server.

### To Reproduce
1. Using this version of BaGet 3d8a219c5116 (docker image latest)
2. Add official Nuget Mirror
3. Restore package Hangfire.SqlServer 1
4. See error:  Microsoft.Data.SqlClient.SqlException (0x80131904): String or binary data would be truncated.

### Expected behavior
Package mirrored.

### Screenshots

```
fail: BaGet.Core.MirrorService[0]
      Failed to mirror package hangfire.sqlserver 1.7.17
Microsoft.EntityFrameworkCore.DbUpdateException: An error occurred while updating the entries. See the inner exception for details.
 ---> Microsoft.Data.SqlClient.SqlException (0x80131904): String or binary data would be truncated.
The statement has been terminated.
   at Microsoft.Data.SqlClient.SqlCommand.<>c.<ExecuteDbDataReaderAsync>b__164_0(Task`1 result)
   at System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke()
   at System.Threading.Tasks.Task.<>c.<.cctor>b__274_0(Object obj)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.EntityFrameworkCore.Storage.RelationalCommand.ExecuteReaderAsync(RelationalCommandParameterObject parameterObject, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Storage.RelationalCommand.ExecuteReaderAsync(RelationalCommandParameterObject parameterObject, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Storage.RelationalCommand.ExecuteReaderAsync(RelationalCommandParameterObject parameterObject, CancellationToken cancellationToken)
[...]

PR #610, opened 2020-11-18
Fix to string truncation error for packages with >4000 char release notes 

dncnkrs:

 * Fixed mismatching column length between database and nuspec

Addresses https://github.com/loic-sharma/BaGet/issues/609

## Issue details

I had this problem with Hangfire, exactly the same package as in #609.

The issue is caused by Hangfire having 6700 character release notes, while BaGet only allocates 4000 character columns for this field. 

The nuspec file reference allows up to 35000 characters in the release notes field: https://docs.microsoft.com/en-us/nuget/reference/nuspec#releasenotes.

To fix this, I've amended the max length of this field to 35000. 

## Caveats etc

I'm no fan of ef migrations, so double check my work here.


Svisstack:
Can we release this after the merge request, so we will be able to ingest that change through entity framework migration after the new baget container startup?

viceice:
closes #590

EdmondFrank:
Good job ! I encountered the same problem. Before that, I only temporarily solved the problem through altering the table field by following statement. 
```sql
ALTER TABLE "Packages" ALTER COLUMN "ReleaseNotes" TYPE text;
```

Svisstack:
Why it’s not permanent?

EdmondFrank:
> Why it’s not permanent?
> […](#)
> Sent from my iPhone
> On 6 May 2021, at 08:56, edmonfrank ***@***.***> wrote: ﻿ Good job ! I encountered the same problem. Before that, I only temporarily solved the problem through altering the table field by following statement. ALTER TABLE "Packages" ALTER COLUMN "ReleaseNotes" TYPE text; — You are receiving this because you commented. Reply to this email directly, view it on GitHub, or unsubscribe.

What I want to express is that when you need deploy the project from zero, this problem still exists. You need manually alter the table again.

splawless (approved):
[](https://github.com/loic-sharma/BaGet/pull/610/files)**_[](https://github.com/loic-sharma/BaGet/pull/610/files)_**

loic-sharma:
Hi @dncnkrs, I apologize for taking so long to review this. I ended up removing the release notes limit in https://github.com/loic-sharma/BaGet/pull/687 as nuget.org has a few packages whose release notes are over 400K! Thanks for your help!

PR #618, opened 2020-12-23
Upgrade to .NET 5

loic-sharma:


BlythMeister:
Propose this is closed in favour of #717 as net5 is out of support in a few weeks.

Issue #224, opened 2019-03-01
support new symbol download url pattern


Ivony:
### Is your feature request related to a problem? Please describe.
baget symbol server is not works in VS2017.
in VS2017, it download symbol file like:
http://baget-host/api/download/symbols/Ne/Newtonsoft.Json.pdb/1F194FBE703D4FCFA949C251C3CCCDE4ffffffff/Newtonsoft.Json.pdb

is't not like

http://baget-host/api/download/symbols/Newtonsoft.Json.pdb/1F194FBE703D4FCFA949C251C3CCCDE4ffffffff/Newtonsoft.Json.pdb

so, it not match baget symbol download route rule.

### Describe the solution you'd like
we can add a new route rule to fix the problem.

```
routes.MapRoute(
  name: Routes.SymbolDownloadRouteName,
  template: "api/download/symbols/{prefix}/{file}/{key}/{file2}",
  defaults: new { controller = "Symbol", action = "Get" });
```

loic-sharma:
Hello, I wasn't able to reproduce this using Visual Studio 2017. How do you create your package? How do you create your snupkg? Is this a .NET package? What framework are you targeting? If possible, please share a project that when packaged reproduces this issue.

loic-sharma:
I'm reluctant from blindly adding this in as there's no symbol store key that accepts a "prefix" as described in this issue. Here are all the documented symbol store routes: https://github.com/dotnet/symstore/blob/master/docs/specs/SSQP_Key_Conventions.md

Ivony:
when VisualStudio download symbol file from symbol server, it's request url like this:
`/api/download/symbols/Sy/System.Security.Principal.pdb/5A62888E7E5D4690A8261A0F66281F23ffffffff/System.Security.Principal.pdb`

I dont know why, but that was the case.
Whether it's Microsoft's server, or other.

Ivony:
![snapshot](https://user-images.githubusercontent.com/521283/58532147-2f3e9500-8217-11e9-808d-cf4319a42cea.png)

pavel-agarkov:
It also does not work in VS 2019

janusk:
In the fiddler capture, we see both `{GUID}ffffffff` as well as `{GUID}1` and it seems that both cases are not considered when doing a lookup of the symbols.

PR #626, opened 2021-02-15
fix: map different key types to the existing directory structure

ahehl:
The symbol files are currently stored in directories with the name {key}ffffffff
In order to support incoming symbol requests with the patterns

{key}ffffffff
{key}1
{key}

the keys have to be mapped to the already existing directory structure

Addresses https://github.com/loic-sharma/BaGet/issues/224


loic-sharma:
I copied your pull request (https://github.com/loic-sharma/BaGet/pull/666) to add a test and merged that copy in. Thanks for the contribution!

PR #630, opened 2021-02-24
Adds support for NuGet V2 upstream package sources

patriksvensson:
This PR adds support for NuGet V2 upstream package sources.

I opted for not rewriting the NuGetClient to use the official API, and instead, I'm just wrapping the existing NuGetClient in an adapter that implements `IMirrorNuGetClient`.

To configure the mirror service to use the NuGet V2 client:

```json
"Mirror": {
  "Enabled": true,
  "PackageSource": "https://www.nuget.org/api/v2/",
  "Legacy": true
},
```

### DIFF
--- a/src/BaGet.Core/BaGet.Core.csproj
+++ b/src/BaGet.Core/BaGet.Core.csproj
@@ -12,7 +12,6 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="$(MicrosoftExtensionsPackageVersion)" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="$(MicrosoftExtensionsPackageVersion)" />
     <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="$(MicrosoftExtensionsPackageVersion)" />
-    <PackageReference Include="NuGet.Packaging" Version="$(NuGetPackageVersion)" />
     <PackageReference Include="NuGet.Protocol" Version="$(NuGetPackageVersion)" />
     <PackageReference Include="System.Reflection.Metadata" Version="1.6.0" />
   </ItemGroup>
--- a/src/BaGet.Core/BaGet.Core.csproj
+++ b/src/BaGet.Core/BaGet.Core.csproj
@@ -13,6 +13,7 @@
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="$(MicrosoftExtensionsPackageVersion)" />
     <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="$(MicrosoftExtensionsPackageVersion)" />
     <PackageReference Include="NuGet.Packaging" Version="$(NuGetPackageVersion)" />
+    <PackageReference Include="NuGet.Protocol" Version="$(NuGetPackageVersion)" />
     <PackageReference Include="System.Reflection.Metadata" Version="1.6.0" />
   </ItemGroup>
 
--- a/src/BaGet.Core/Configuration/MirrorOptions.cs
+++ b/src/BaGet.Core/Configuration/MirrorOptions.cs
@@ -18,6 +18,11 @@ namespace BaGet.Core
         public Uri PackageSource { get; set; }
 
         /// <summary>
+        /// Whether or not the package source is a v2 package source feed.
+        /// </summary>
+        public bool Legacy { get; set; }
+
+        /// <summary>
         /// The time before a download from the package source times out.
         /// </summary>
         [Range(0, int.MaxValue)]
--- a/src/BaGet.Core/Mirror/Clients/MirrorV2Client.cs
+++ b/src/BaGet.Core/Mirror/Clients/MirrorV2Client.cs
@@ -1,0 +1,114 @@
+using BaGet.Protocol.Models;
+using Microsoft.Extensions.Options;
+using NuGet.Common;
+using NuGet.Configuration;
+using NuGet.Protocol;
+using NuGet.Protocol.Core.Types;
+using NuGet.Versioning;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace BaGet.Core
+{
+    internal sealed class MirrorV2Client : IMirrorNuGetClient
+    {
+        private readonly ILogger _logger;
+        private readonly SourceCacheContext _cache;
+        private readonly SourceRepository _repository;
+
+        public MirrorV2Client(IOptionsSnapshot<MirrorOptions> options)
+        {
+            if (options is null)
+            {
+                throw new ArgumentNullException(nameof(options));
+            }
+
+            if (options.Value?.PackageSource?.AbsolutePath == null)
+            {
+                throw new ArgumentException("No mirror package source has been set.");
+            }
+
+            _logger = NullLogger.Instance;
+            _cache = new SourceCacheContext();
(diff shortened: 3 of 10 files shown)


src/BaGet.Core/BaGet.Core.csproj
```diff
@@ -13,6 +13,7 @@
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="$(MicrosoftExtensionsPackageVersion)" />
     <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="$(MicrosoftExtensionsPackageVersion)" />
     <PackageReference Include="NuGet.Packaging" Version="$(NuGetPackageVersion)" />
+    <PackageReference Include="NuGet.Protocol" Version="$(NuGetPackageVersion)" />
```
loic-sharma, line 16: `NuGet.Protocol` depends on `NuGet.Packaging`, you should be able to delete line 15.

image

src/BaGet.Core/Mirror/Clients/MirrorV3Client.cs
```diff
@@ -0,0 +1,36 @@
...
+        private readonly NuGetClient _client;
+
+        public MirrorV3Client(NuGetClient client)
+        {
+            _client = client ?? throw new ArgumentNullException(nameof(client));
+        }
+
+        public Task<Stream> DownloadPackageAsync(string id, NuGetVersion version, CancellationToken cancellationToken)
+        {
+            return _client.DownloadPackageAsync(id, version, cancellationToken);
```
loic-sharma, line 23: Please `await` these tasks. See: https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#prefer-asyncawait-over-directly-returning-task

src/BaGet.Core/Mirror/MirrorNuGetClientAdapter.cs
```diff
@@ -0,0 +1,8 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BaGet.Core.Mirror
+{
```
loic-sharma, line 6: Looks like this file should be deleted :)

src/BaGet.Core/Configuration/MirrorOptions.cs
```diff
@@ -17,6 +17,11 @@ public class MirrorOptions : IValidatableObject
         /// </summary>
         public Uri PackageSource { get; set; }
```
loic-sharma, line 18: Hey this is a tiny bit of feature creep, but I think you can leverage the options validation here by adding attributes on the `PackageSource` property:

```diff
/// <summary>
/// The v3 index that will be mirrored.
/// </summary>
[RequiredIf(nameof(Enabled))]
[RequiredIf(nameof(Legacy))]
public Uri PackageSource { get; set; }

File: src/BaGet.Core/Mirror/Clients/MirrorV2Client.cs [hunk 5]
```diff
@@ -0,0 +1,114 @@
...
+        public MirrorV2Client(IOptionsSnapshot<MirrorOptions> options)
+        {
+            if (options is null)
+            {
+                throw new ArgumentNullException(nameof(options));
+            }
+
+            if (options.Value?.PackageSource?.AbsolutePath == null)
+            {
+                throw new ArgumentException("No mirror package source has been set.");
```
loic-sharma, line 32: Ideally this should be done at startup by ASP.NET Core's options validation. See my comment up above: https://github.com/loic-sharma/BaGet/pull/630/files#r582447197

If that idea works, you should be able to remove this check.

loic-sharma:
This is _amazing_. I really like your `IMirrorNuGetClient` abstraction, good idea there! This change will make it much easier to add support for private upstreams in the future too.

I left some minor comments, I'll merge this in once you've addressed them. Thanks!

src/BaGet.Core/Configuration/MirrorOptions.cs
patriksvensson: Excellent idea. However when trying this it complains that only one `IfRequired` attribute can be used.

As I workaround, we can do the following, that will work but will show the error message `Invalid 'Mirror' configs: 'PackageSource' is required because 'EnabledOrLegacy' has a value of 'True'.` which does not reflect the settings properly, but maybe this is OK?

```
/// <summary>
/// If true, packages that aren't found locally will be indexed
/// using the upstream source.
/// </summary>
public bool Enabled { get; set; }

/// <summary>
/// The v3 index that will be mirrored.
/// </summary>
[RequiredIf(nameof(EnabledOrLegacy), true)]
public Uri PackageSource { get; set; }

/// <summary>
/// Whether or not the package source is a v2 package source feed.
/// </summary>
public bool Legacy { get; set; }

[NotMapped]
public bool EnabledOrLegacy => Enabled || Legacy;
```

src/BaGet.Core/BaGet.Core.csproj
patriksvensson: Done!

src/BaGet.Core/Mirror/MirrorNuGetClientAdapter.cs
patriksvensson: Done!

src/BaGet.Core/Mirror/Clients/MirrorV3Client.cs
patriksvensson: Done!

patriksvensson:
@loic-sharma I've fixed all points except the validation part (see my comment). I'm happy to implement the changes if you want to. Btw, do you want me to squash the commits?

> P.S. Did you have any pain points when contributing to BaGet? Let me know if you have any feedback, would love to improve the experience however I can :)

No pain points! I found the code base intuitive and easy to follow.  Of course, there are aspects of how NuGet works that I'm not familiar with, but I think it's unavoidable to have to learn that.

One thing that I was a bit confused about was a test (`RegistrationBuilderTests.TheRegistrationIndexResponseIsSortedByVersion`) that fails consistently for me but the CI checks seem to be green. Not sure if this is something culture-related since I'm using a computer with Swedish regional settings.

src/BaGet.Core/Configuration/MirrorOptions.cs
loic-sharma: Ah good catch! It looks like you can support multiple attribute usages if we:

1. Set the attribute's `AttributeUsageAttribute.AllowMultiple` to true
1. Override the `TypeId` property (see [this](https://stackoverflow.com/questions/553540/how-to-create-duplicate-allowed-attributes))

For an example of a validation attribute that supports multiple uses, see [`CustomValidationAttribute`](https://github.com/dotnet/runtime/blob/b51270d64f833b87d33639341392eb8f35442efa/src/libraries/System.ComponentModel.Annotations/src/System/ComponentModel/DataAnnotations/CustomValidationAttribute.cs). I'll do this in a follow-up, I've blocked your changes for a little too long already!

loic-sharma:
Thanks for the contribution, this is awesome! I'll work on publishing a new version of BaGet.

> I've fixed all points except the validation part (see my comment). I'm happy to implement the changes if you want to.

That's alright, I'm very happy with your implementation already. I'd rather release the feature now, we can save incremental improvements for later :)

> Btw, do you want me to squash the commits?

Feel free to do whatever works best for you! I squash the commits at the very end when I merge.

> One thing that I was a bit confused about was a test (`RegistrationBuilderTests.TheRegistrationIndexResponseIsSortedByVersion`) that fails consistently for me

Thanks for the heads up, I'll take a look at that!

Issue #633, opened 2021-03-03
Publish version as docker tags


viceice:
### Is your feature request related to a problem? Please describe.
I like to have [renovate](/renovatebot/renovate) to be able to suggest updates with release notes.
Renovate is able to match the docker tag to a github release / git tag.

### Describe the solution you'd like
Please publish the corresponding github releases as docker tag and add the `org.opencontainers.image.source="https://github.com/loic-sharma/BaGet"` label to the image.

### Describe alternatives you've considered
No alternate 😕 

### Additional context
That way we can use a specific docker tag to reference and let renovate suggest updates to newer versions Instead of relying on latest digest and manually find the changes happened.

loic-sharma:
Excellent suggestion! Versions will be tagged starting from 0.4.0-preview1 and on! 😎

PR #634, opened 2021-03-03
feat: add docker source label

viceice:
Add docker source label, so the image points back to this repo.

Addresses #633 partly


loic-sharma:
Thanks for the contribution!

Issue #621, opened 2021-01-19
Support alternate S3 storage provider


viceice:
### Is your feature request related to a problem? Please describe.
I Like to host baget internally using postgresql and minio (s3 compatible)

### Describe the solution you'd like
I Like to oerride the s3 url and use path style buckets.

### Describe alternatives you've considered
Using a shared filesystem. Is this possible?

viceice:
Looks like it's easy.

ref: https://docs.min.io/docs/how-to-use-aws-sdk-for-net-with-minio-server.html

I can send a pr

Issue #551, opened 2020-07-08
Overriding endpoints in the AWS S3 storage


thomas-illiet:
Hello Loic :)

Can you implement the overriding endpoint service url in the s3 storage.

I need this to use internal storage in my company.

Kind regards
Thomas

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

loic-sharma:
I'm not very familiar with this. From reading [this guide](https://aws.amazon.com/blogs/developer/overriding-endpoints-in-the-aws-sdk-for-net/), it looks like this would mean overriding the `AmazonS3Config.ServiceURL` property?

This should be pretty to do here: https://github.com/loic-sharma/BaGet/blob/68e99bf416cb547d5454d3d2b54b27e07be48091/src/BaGet.Aws/AwsApplicationExtensions.cs#L34-L37

I'd be happy to accept a pull request

thomas-illiet:
I will try to create a pull request next week 🤪

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Issue #271, opened 2019-05-17
Configure local AWS (S3) service


DTeuchert:
### Is your feature request related to a problem? Please describe.
I like to use a on-premise AWS S3 cloud service, which is running in my network locally.

### Describe the solution you'd like
The current configuration of amazon S3 client only use the `RegionEndpoint` property.
This field needs to get optional and a new property `ServiceURL` need to be definied.

Optional also the other [AmazonS3Config properties](https://docs.aws.amazon.com/sdkfornet1/latest/apidocs/html/T_Amazon_S3_AmazonS3Config.htm) could be implemented.

PR #644, opened 2021-04-28
Minio (S3-Compatible) storage backend

mattj23:
This PR adds a `IStorageService` to BaGet that uses Minio, an open-source S3 compatible storage provider popular for self-hosted and internal object storage.  Furthermore the [.NET Minio Client Library](https://github.com/minio/minio-dotnet) used in this PR supports any S3 compatible provider and so can also be used directly for AWS as well as Minio.

This addresses #621, #551, and #271 (which resulted in PR #272 but never got merged).

I copied the layout of the AWS provider, making changes only for the specific Minio client libraries and configuration options.


thomas-illiet:
Great feature! Well played.
I need this to use the product 🤤

PR #654, opened 2021-05-28
Configuration of IISServerOptions from appsettings now occurs after defaults are configured

### LINKED ISSUE #653: IISServerOptions.MaxRequestBodySize from appsettings is ignored
### Describe the bug
Setting IISServerOptions.MaxRequestBodySize in the appsettings file has no effect on the application. The default value of `262144000` is always used.

### To Reproduce
1. Set `IISServerOptions.MaxRequestBodySize` in appsettings.json to `524288000`
2. Run BaGet via IIS.
3. Push a NuGet package that is greater than 250MB
4. The push will be rejected with a 413 response because the body is too large.

### Expected behavior
If `IISServerOptions.MaxRequestBodySize` is specified in appsettings.json, then that value should be used.

reduckted:
Fixes #653.

The IISServerOptions were being populated from appsettings, but were then being overridden with the default configuration. This caused the `MaxRequestBodySize` option to _always_ be `262144000`, regardless of what was configured in appsetings.

### DIFF
--- a/src/BaGet/Startup.cs
+++ b/src/BaGet/Startup.cs
@@ -24,8 +24,6 @@ namespace BaGet
 
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddBaGetOptions<IISServerOptions>(nameof(IISServerOptions));
-
             // TODO: Ideally we'd use:
             //
             //       services.ConfigureOptions<ConfigureBaGetOptions>();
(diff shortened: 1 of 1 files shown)


Highsight:
Can confirm that this change has fixed my IIS Payload Too Large issues. I hope it gets merged in soon!

reduckted:
@Highsight What I've done is downloaded the "drop" artifact from the build and used that to install BaGet.
https://dev.azure.com/sharml/BaGet/_build/results?buildId=1608&view=artifacts&pathAsName=false&type=publishedArtifacts

loic-sharma:

cenit:
@reduckted hi! first of all, thanks for the PR
Even with this fix, I am unable to go beyond the 2GB limit. 
I tried to put 8GB for example in the appsettings.json, then I tried to push a ~4GB nuget package. After the first 2GB, i receive an `InternalServerError` message and the push action restarts from the beginning, in a loop.
Is it a hard limit of IIS? Or is there any way to upload very big nuget packages?

PR #667, opened 2021-07-14
Add support for MiniO storage backend

alfeg:
This PR is yet another attempt to address  #621, #551, and #271

Added additional settings to `S3StorageOptions`:
- `ForcePathStyle = false` setting to true will make AWSSDK.net compatible with MiniO installations
- `ServiceUrl` - ability to override S3 service location
- `UseHttp` - ability to use HTTP protocol for `ServiceUrl`

All those settings are real settings from actual `AmazonS3Config` from official AWSSDK library

There is already another PR's to address those issues: https://github.com/loic-sharma/BaGet/pull/644 and https://github.com/loic-sharma/BaGet/pull/272

My PR is much less intrusive, only adding ability to pass settings that are already exists in AwsSdk library

### DIFF
--- a/docs/installation/aws.md
+++ b/docs/installation/aws.md
@@ -22,12 +22,17 @@ Update the `appsettings.json` file:
         "Region": "us-west-1",
         "Bucket": "foo",
         "AccessKey": "",
-        "SecretKey": ""
+        "SecretKey": "",
+        "ServiceUrl": "",
+        "UseHttp": false,
+        "ForcePathStyle": false
     },
 
     ...
 }
 ```
+
+To support MiniO as S3 compatible storage backend provide MiniO installation url in `ServiceUrl`, set `ForcePathStyle` to true.
 
 ### Amazon RDS
 
--- a/docs/installation/aws.md
+++ b/docs/installation/aws.md
@@ -32,7 +32,25 @@ Update the `appsettings.json` file:
 }
 ```
 
-To support MiniO as S3 compatible storage backend provide MiniO installation url in `ServiceUrl`, set `ForcePathStyle` to true.
+To use MinIO as an S3 compatible storage backend, provide the MinIO installation URL in `ServiceUrl` and set `ForcePathStyle` to `true`:
+
+```json
+{
+    ...
+
+    "Storage": {
+        "Type": "AwsS3",
+        "Region": "us-west-1",
+        "Bucket": "foo",
+        "AccessKey": "",
+        "SecretKey": "",
+        "ServiceUrl": "",
+        "ForcePathStyle": false,
+        "UseHttp": false
+    },
+
+    ...
+}
 
 ### Amazon RDS
 
--- a/src/BaGet.Aws/AwsApplicationExtensions.cs
+++ b/src/BaGet.Aws/AwsApplicationExtensions.cs
@@ -31,10 +31,23 @@ namespace BaGet
             {
                 var options = provider.GetRequiredService<IOptions<S3StorageOptions>>().Value;
 
+                if (options.ForcePathStyle)
+                {
+                    // Required to support presigned urls generation for MiniO configured to us-east-1 region
+                    AWSConfigsS3.UseSignatureVersion4 = true; 
+                }
+
                 var config = new AmazonS3Config
                 {
-                    RegionEndpoint = RegionEndpoint.GetBySystemName(options.Region)
+                    RegionEndpoint = RegionEndpoint.GetBySystemName(options.Region),
+                    ForcePathStyle = options.ForcePathStyle,
+                    UseHttp = options.UseHttp
                 };
+
+                if (!string.IsNullOrWhiteSpace(options.ServiceUrl))
+                {
+                    config.ServiceURL = options.ServiceUrl;
+                }
 
                 if (options.UseInstanceProfile)
                 {
--- a/src/BaGet.Aws/S3StorageOptions.cs
+++ b/src/BaGet.Aws/S3StorageOptions.cs
@@ -22,5 +22,11 @@ namespace BaGet.Aws
         public bool UseInstanceProfile { get; set; }
 
         public string AssumeRoleArn { get; set; }
+
+        public bool ForcePathStyle { get; set; } = false;
+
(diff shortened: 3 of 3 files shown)


viceice:
@loic-sharma Can you please add this? I like BaGet but would like to use MinIO as storage backend in my kubernetes cluster instead of nfs.

docs/installation/aws.md
```diff
@@ -22,13 +22,18 @@ Update the `appsettings.json` file:
         "Region": "us-west-1",
         "Bucket": "foo",
         "AccessKey": "",
-        "SecretKey": ""
+        "SecretKey": "",
```
loic-sharma, line 25: Folks should use `AccessKey` instead of `SecretKey` right?

```suggestion
```

src/BaGet.Aws/S3StorageOptions.cs
```diff
@@ -22,5 +22,11 @@ public class S3StorageOptions
         public bool UseInstanceProfile { get; set; }
 
         public string AssumeRoleArn { get; set; }
+
+        public bool ForcePathStyle { get; set; } = false;
```
loic-sharma, line 26: ASP.NET Core will default this to false automatically, no need to be explicit here:

```suggestion
        public bool ForcePathStyle { get; set; }
```

docs/installation/aws.md
```diff
@@ -22,13 +22,18 @@ Update the `appsettings.json` file:
...
+        "ServiceUrl": "",
+        "UseHttp": false,
+        "ForcePathStyle": false
     },
 
     ...
 }
 ```
 
+To support MiniO as S3 compatible storage backend provide MiniO installation url in `ServiceUrl`, set `ForcePathStyle` to true.
```
loic-sharma, line 35:
```suggestion
To support MinIO as S3 compatible storage backend provide MinIO installation url in `ServiceUrl`, set `ForcePathStyle` to true.
```

docs/installation/aws.md
```diff
@@ -22,13 +22,18 @@ Update the `appsettings.json` file:
         "Region": "us-west-1",
         "Bucket": "foo",
         "AccessKey": "",
-        "SecretKey": ""
+        "SecretKey": "",
+        "ServiceUrl": "",
+        "UseHttp": false,
+        "ForcePathStyle": false
```
loic-sharma, line 28: It'd be less confusing to folks if we separate the AWS S3 and MinIO samples:

```suggestion
```

docs/installation/aws.md
loic-sharma: Oops, just kidding this is right.

loic-sharma:
Hello @alfeg, thank you for the contribution! Your changes look good. I suggested some minor tweaks but I've never used MinIO. Could you let me know what you think?

@viceice Feel free to jump in too if you have comments!

src/BaGet.Aws/S3StorageOptions.cs
```diff
@@ -22,5 +22,11 @@ public class S3StorageOptions
...
+
+        public string ServiceUrl { get; set; }
```
loic-sharma, line 28: Could we add the data annotation attributes that @DTeuchert had used in https://github.com/loic-sharma/BaGet/pull/272?

```diff
-       [Required]
+       [RequiredIf(nameof(ServiceUrl), null)]
        public string Region { get; set; }

+       [RequiredIf(nameof(Region), null)]
        public string ServiceUrl { get; set; }
```

src/BaGet.Aws/AwsApplicationExtensions.cs
```diff
@@ -31,11 +31,24 @@ public static BaGetApplication AddAwsS3Storage(this BaGetApplication app)
...
+                if (options.ForcePathStyle)
+                {
+                    // Required to support presigned urls generation for MiniO configured to us-east-1 region
+                    AWSConfigsS3.UseSignatureVersion4 = true; 
+                }
+
                 var config = new AmazonS3Config
                 {
-                    RegionEndpoint = RegionEndpoint.GetBySystemName(options.Region)
+                    RegionEndpoint = RegionEndpoint.GetBySystemName(options.Region),
```
loic-sharma, line 42: From my understanding the region can be set using the service URL. If so should we follow @DTeuchert's pattern in https://github.com/loic-sharma/BaGet/pull/272:

```suggestion
                    RegionEndpoint = (options.Region != null)
                        ? RegionEndpoint.GetBySystemName(options.Region)
                        : null,
                    ServiceURL = options.ServiceUrl
```

alfeg:
> Hello @alfeg, thank you for the contribution! Your changes look good. I suggested some minor tweaks but I've never used MinIO. Could you let me know what you think?

Hello,

I'm OK with all proposed changes, but cannot test them in any way for next two weeks. I'm on vacation without laptop and with poor internet :)

docs/installation/aws.md
viceice: Sound good

src/BaGet.Aws/AwsApplicationExtensions.cs
viceice: I think the AWS SDK requires a region independent from service url

viceice (commented):
Sounds good

src/BaGet.Aws/S3StorageOptions.cs
DTeuchert: Yes in my understanding we need a service url for a private / on premise solutions  or a region.

DTeuchert:
On Monday I can test this request with a ceph cluster as s3 backend.

krmxd:
are there any plans on following up on this ?

viceice:
you should check 
- https://github.com/bagetter/BaGetter

pishangujeniya:
@loic-sharma Any plans on this?

loic-sharma:
No. I'd consider checking out BaGetter instead, it is an active fork of BaGet.

pishangujeniya:
> No. I'd consider checking out BaGetter instead, it is an active fork of BaGet.

that is also not having support of MinIO.

PR #689, opened 2021-09-20
fix: use new dotnet images

viceice:

 * dotnet image names have changed

https://hub.docker.com/_/microsoft-dotnet-aspnet
https://hub.docker.com/_/microsoft-dotnet-sdk


loic-sharma:
Good stuff, I wasn't aware of this change. Thanks for the contribution! :)

PR #737, opened 2022-06-29
Fix unclosed XML tag in PackageReference syntax

nloum:
When you want to add a dependency from a BaGet server, you have the option to copy the `<PackageReference>` element that would go in the .csproj file. However the element you copy is malformed: it is unclosed. E.g., here's what you copy:

```
<PackageReference Include="Mypackage" Version="1.2.3">
```

What you should copy is this:

```
<PackageReference Include="Mypackage" Version="1.2.3"/>
```

Notice the `/>` at the end. The backslash was missing. This PR fixes that.


loic-sharma:
Thanks for the contribution!

Issue #761, opened 2023-01-25
.NET 6 BaGet


mbaksh01:
Would it make send to upgrade this project to .NET 6 as .NET Core 3.1 is now out of support?

dansiegel:
Since it seems @loic-sharma doesn't seem to have time to update this. For those who need net6+ support you might check out a [fork of BaGet](https://github.com/avantipoint/avantipoint.packages) that I've been maintaining which also has some additional functionality for providing your own custom Authentication Handler's. It's currently being cross compiled for net6.0, net7.0 and net8.0 with net7.0 and net8.0 converted to use minimal APIs that also have the additional metadata for using Swagger.

PR #775, opened 2023-07-05
Net6

sunny1028:
主版本升级为.NET7
并且升级了绝大部分相关依赖版本
目前为止运行状态良好

如果需要可以自行构建Releases

Addresses https://github.com/loic-sharma/BaGet/issues/761


jkone27:
yes pls NET6 LTS

src/BaGet.Aliyun/BaGet.Aliyun.csproj
```diff
@@ -2,15 +2,25 @@
...
     <PackageTags>NuGet;Alibaba;Cloud</PackageTags>
     <Description>The libraries to host BaGet on Alibaba Cloud (Aliyun).</Description>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <LangVersion>10.0</LangVersion>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
+    <LangVersion>10.0</LangVersion>
```
Regenhardt: Isn't the langversion coupled to the framework now anyway, do we really need this?

src/BaGet.Aliyun/BaGet.Aliyun.csproj
```diff
@@ -2,15 +2,25 @@
...
-    <PackageReference Include="Aliyun.OSS.SDK.NetCore" Version="2.9.1" />
+    <PackageReference Include="Aliyun.OSS.SDK.NetCore" Version="2.13.0" />
   </ItemGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\BaGet.Core\BaGet.Core.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Update="Microsoft.SourceLink.GitHub" Version="1.1.1" />
```
Regenhardt: Have you tested this in a docker container? I recall sourcelink being complicated when trying to put it in a docker container. Also do we really need sourcelink for something that is mainly a standalone application?

src/BaGet/appsettings.json
```diff
@@ -30,11 +29,11 @@
   //"Kestrel": {
   //  "Endpoints": {
   //    "Http": {
-  //      "Url": "http://localhost:8080"
+  //      "Url": "http://*:11050"
```
Regenhardt: It's nice with the asterisk, but why port 11050?

Regenhardt (changes requested):

BaGet.sln
```diff
@@ -1,26 +1,18 @@
...
+# Visual Studio Version 17
+VisualStudioVersion = 17.3.32901.215
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "BaGet", "src\BaGet\BaGet.csproj", "{284366CB-C68F-473E-908A-50A382616AE0}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "BaGet.Core", "src\BaGet.Core\BaGet.Core.csproj", "{FFFACD28-C300-4046-BCFE-4A7899E88EA3}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "BaGet.Azure", "src\BaGet.Azure\BaGet.Azure.csproj", "{716C970D-9614-4265-AC92-57E8B227B98E}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "BaGet.Core.Tests", "tests\BaGet.Core.Tests\BaGet.Core.Tests.csproj", "{89AB1AE2-6CAA-4809-8B74-D78CBE00B049}"
```
Regenhardt, line 12: No more tests?

joestr:
After enabling the tests I get following error:
`The UTC time represented when the offset is applied must be between year 0 and 10,000`

I fixed it by Replacing occourances of `DateTime` with `DateTimeOffset`

Regenhardt:
The only way I can imagine this error happening is if there's a datetime of 0001-01-01 and then the offset pushes it below that date. I'm kinda sick today so I won't check now, but if you push it with tests enabled I can checkout your branch tomorrow and see if I can find something causing that. I don't think there should be a datetime of 0001-01-01 anywhere so this might be important.

joestr:
I just had time to publish the changes again @Regenhardt. They are [here](https://github.com/joestr/BaGet-net6-pr-fix/tree/net6)

joestr:
Fixed the last Unit Test in #786

PR #777, opened 2023-07-10
Mysql storage

Shamus03:
Add a new storage type option that stores package contents in MySql
along with the package metadata.

Though MySql is not the best for storing large file contents, this is an
alternative to storing package contents in the file system, which should be
better for highly-available/distributed use cases.

### DIFF
--- a/src/BaGet.Core/Extensions/DependencyInjectionExtensions.cs
+++ b/src/BaGet.Core/Extensions/DependencyInjectionExtensions.cs
@@ -100,6 +100,7 @@ namespace BaGet.Core
 
             services.TryAddTransient<DatabaseSearchService>();
             services.TryAddTransient<FileStorageService>();
+            services.TryAddTransient<DatabaseStorageService>();
             services.TryAddTransient<PackageService>();
             services.TryAddTransient<V2UpstreamClient>();
             services.TryAddTransient<V3UpstreamClient>();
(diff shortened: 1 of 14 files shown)


Shamus03:
More information about our use case.  We run BaGet in multiple datacenters pointing to a MySql database.  Since BaGet only supports saving package contents to the file system, we use a distributed file system to store those documents.  However, distributed file systems are not perfect and sparing you the details, we think it would be better for us to store the package contents in our MySql database along with the package metadata so everything is in one place.  I think this will not be an uncommon thought.

src/BaGet.Core/Extensions/DependencyInjectionExtensions.cs
```diff
@@ -133,6 +134,11 @@ private static void AddDefaultProviders(this IServiceCollection services)
                     return provider.GetRequiredService<FileStorageService>();
                 }
 
+                if (configuration.HasStorageType("mysql"))
```
Shamus03, line 137: I realized this should perhaps be "database" not "mysql".  What do you think @loic-sharma?

src/BaGet.Database.MySql/Migrations/20230710152239_CreatePackageContentsTable.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            migrationBuilder.AlterColumn<DateTime>(
+                name: "RowVersion",
+                table: "Packages",
+                rowVersion: true,
+                nullable: true,
+                oldClrType: typeof(DateTime),
+                oldType: "timestamp(6)",
+                oldNullable: true)
+                .Annotation("MySql:ValueGenerationStrategy", MySqlValueGenerationStrategy.ComputedColumn);
```
Shamus03, line 19: I was tempted to remove this but I see it is in every other migration file.  Do you have a preference?

Shamus03:
This repo seems abandoned.  Anyone else please feel free to steal this code for another project.

